PR #13, opened 2022-09-09
6 - Set up sentry integration

nozomione:
## Issue Number
#6

Setup the [Sentry Next.js SDK](https://docs.sentry.io/platforms/javascript/guides/nextjs/) for error tracking.

Updates include:
- [x] Install Sentry SDK
- [x] Creat a new project for refinebio-web in our Sentry account - **Completed by David**
- [x] Set up the basic [configuration files](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#create-initialization-config-files) for Sentry 
- [x] Update Next.js configuration to incorporate Sentry
- [x] Define the sentry DSN/ENV environmental variables in local and Vercel  - **use the Sentry's Client Key provided by David**
- [x] Enable source map settings in Next.js(enabled by default via Sentry SDK)
- [x] Add an [error boundary](https://docs.sentry.io/platforms/javascript/guides/react/components/errorboundary/) in the codebase(incorporate the existing pattern for consistency - set error boundary at the page level)
- [x] Test and verified the sentry integration in our Sentry account - **with David**

Thank you, David!

## Checklist


nozomione:
@davidsmejia Could you please confirm the Sentry integration(since no account access on my end)? Thank you, David!

.env.development
```diff
@@ -0,0 +1,2 @@
+STAGE_SENTRY_DSN=https://758146f240094216826ede840ab34857@o7983.ingest.sentry.io/6703919
+STAGE_SENTRY_ENV=client-local-development
```
davidsmejia, line 2:
```suggestion
STAGE_SENTRY_ENV=local
```

these will all be for refinebio-web which is a client so we can just use local, preview, production

src/pages/_app.js
```diff
@@ -1,6 +1,15 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
davidsmejia, line 1: can we move this rule to where the exception is

sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=process.env.SENTRY_ORG
+defaults.project=process.env.SENTRY_PROJECT
+cli.executable=../../.npm/_npx/a8388072043b4cbc/node_modules/@sentry/cli/bin/sentry-cli
```
davidsmejia, line 4: I'm not convinced this looks correct. I think this will work on your local env but not elsewhere. I will try this locally to confirm

nozomione: Yes, we're not utilizing this file currently, thus I'll remove it from the codebase, thank you!

nozomione:
@davidsmejia 
> these will all be for refinebio-web which is a client so we can just use local, preview, production

For the environment variable names, are there any rules we should apply to them based on the above.?

sentry.client.config.js
```diff
@@ -0,0 +1,10 @@
+// This file configures the initialization of Sentry on the browser:
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN
+
+Sentry.init({
+  dsn: SENTRY_DSN
```
davidsmejia, line 9: i think this needs the environment here, same with the other config file

davidsmejia, line 9: add env here

sentry.server.config.js
```diff
@@ -0,0 +1,7 @@
+import * as Sentry from '@sentry/nextjs'
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN
+
+Sentry.init({
+  dsn: SENTRY_DSN
```
davidsmejia, line 6: same here

davidsmejia (approved):
Please update the files with comments but looks good!

Issue #26, opened 2022-10-06
Create the global components and responsive UI rules of the application 


nozomione:
### Context
We need to define the global components and responsive UI rules of the application.

### Problem or idea
Create the necessary global components and responsive UI rules, which includes the following:
- Layout
- Header
- Footer
- Sidebar
- breakpoints and style rules etc

nozomione:
@davidsmejia This PR will be created based on the branch `nozomione/20-setup-storybook` to use the latest docker setting and Storybook features(accessibility testing etc), thank you!

Issue #30, opened 2022-10-24
Implement About Page


davidsmejia:
### Context

Add the necessary components and implement them on the about page.

nozomione:
@davidsmejia This PR will be created based on the branch ` nozomione/26-create-global-components-for-app-structure`, for it's a dependent of the issue [#26](https://app.zenhub.com/workspaces/engineering-5c92300a9d1bd30b8d3101f2/issues/alexslemonade/refinebio-web/31), thank you!

Issue #29, opened 2022-10-24
Homepage (Without Cached Data)


davidsmejia:
Implement the components required to build the homepage and implement the homepage.

Using:
- Layout
- Header
- Footer

nozomione:
@davidsmejia This PR will be created based on the branch `nozomione/20-setup-storybook` to use the latest docker setting and Storybook features(accessibility testing etc), thank you!

PR #31, opened 2022-10-31
26 - Create the global components and responsive UI rules of the application

nozomione:
## Issue Number

#26 

(**NOTE:** This PR was created based on the branch `nozomione/20-setup-storybook` to use the latest docker setting etc.)

Defined the global components and responsive UI rules of the application. 

Created the necessary global components and the basic responsive UI rules, which includes the following:

**Components:**
- `Layout`
- `Header`
- `Footer`

(shared components)
- `Anchor`
- `Fixed Container`
- `SrOnly`

**Themes** 
- breakpoints and some adjustments made in `/themes/` 
- `styles/GlobalStyle`  (added reset for a mobile)

**Hook**
-  `hooks/useResponsive`(to utilize the Grommet `ResponsiveContext`)

**Stories**
- `Header`
- `Footer` 

(5 svg image files)

- [x] UI review w/ @dvenprasad - please review the UI [here](https://refinebio-web-git-nozomione-26-create-global-compon-47c2a0-ccdl.vercel.app/), thank you!

(After passing the UI review, @davidsmejia will be added to the reviewers.)

## Types of changes

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist


dvenprasad (changes requested):
Hey most of it looks good! I have a few minor notes:

- For the footer, let's add a drop shadow so it stands out against white background. 
  This is what we currently have on refine.bio: ` box-shadow: 0 3px 20px 0 rgba(0,0,0,.1);`

- On the footer social icons, can we have the Twitter turn #55acee on hover and the Github turn black on hover

-  The responsive navigation looks great! There was just one screen where we could add a bit more padding between the logo and `Search`. it's for the iphone SE sized screen

Screen Shot 2022-11-07 at 2 37 59 PM

nozomione:
@dvenprasad, the UI is ready for your review, thank you!

dvenprasad (approved):
LGTM! 🐐 🐐 🐐

nozomione:
@davidsmejia, the following changes were made in the latest PRs (please reference [#29](https://github.com/AlexsLemonade/refinebio-web/pull/32)/[#30](https://github.com/AlexsLemonade/refinebio-web/pull/33)):

- Optimized the `Header/*` components
- Optimized the `Footer/*` component
   - Flatten the folder and deleted the `config.js`
   - Added `src/config` directory and created [`contributors.js`](https://github.com/AlexsLemonade/refinebio-web/pull/33/files#diff-38087c66d7d514666f839e5d916cc4c9b5e0e0a3c207af6b11cd07749815e9df) and [`links.js`](https://github.com/AlexsLemonade/refinebio-web/pull/33/files#diff-f43717887fed42467fbfa19b7c175a160ff16c72b793a26986a7c8d2cafcc1fe) (shared with other components, `Home`/`About`)
- Updated the `Anchor` components to utilize the next `Link`component for routes
- Removed `styled-component` and a wrapper from the following components:
   - `Button`, `Col`, `Logo`, `SrOnly` 
- Created a helper for `edgeSize` to auto-generate key/value pairs in `themes/global.js` 
- Remove `themes/variables.js` and updated the codebase accordingly.

Thank you, David!

src/components/Header/Logo.js
```diff
@@ -1,22 +1,29 @@
 import { Box } from 'grommet'
+import { Anchor } from 'components/shared/Anchor'
+import { SrOnly } from 'components/shared/SrOnly'
 import styled, { css } from 'styled-components'
 import { LogoSvg } from '../../images/logo.svg'
 
 // NOTE: Set the prop 'light' to true for the dark background
 
 const Wrapper = styled(Box)`
```
davidsmejia, line 9: I don't think we need to use a wrapper here. `LogoAnchor` makes more sense to me.

src/components/shared/Col.js
```diff
@@ -0,0 +1,34 @@
+import { useResponsive } from 'hooks/useResponsive'
+import { Box } from 'grommet'
+import styled, { css } from 'styled-components'
+
+const Wrapper = styled(Box)`
+  ${({ flexValue }) => css`
+    flex: ${flexValue};
```
davidsmejia, line 7: Couple things here, I think the name `ColBox` makes more sense here. Also, there are attributes you can use to manage the flex type.

src/components/shared/Col.js
```diff
@@ -0,0 +1,34 @@
...
+  `}
+
+  ${({ viewport }) =>
+    viewport === 'small' &&
+    css`
+      flex: 1 1 auto;
+      + div {
```
davidsmejia, line 14: We shouldn't be applying styles to a sibling div here. There are other ways to accomplish what this does that allows for more transparency in how / when styles are applied.

src/components/shared/Layer.js
```diff
@@ -0,0 +1,13 @@
+import { Layer as GrommetLayer } from 'grommet'
+import { useResponsive } from 'hooks/useResponsive'
+
+export const Layer = ({ position = '', show = false, children }) => {
+  const { viewport } = useResponsive()
+  if (viewport !== 'small') return children
+  return (
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    <>{show && <GrommetLayer position={position}>{children}</GrommetLayer>}</>
```
davidsmejia, line 9: This rule exception doesn't need to be applied here. Your component can have two returns, if `show` is false return before the markup

src/components/shared/SrOnly.js
```diff
@@ -0,0 +1,18 @@
+import { Text } from 'grommet'
+import styled from 'styled-components'
+
+const Wrapper = styled(Text)`
```
davidsmejia, line 4: You can rename this to be more descriptive like `HiddenText`

davidsmejia (changes requested):
Some changes for consistency and clarity here. Looking good!

src/components/Footer/index.js
```diff
@@ -0,0 +1,185 @@
...
+      justify="center"
+      fill
+      gap="0"
+      pad={{ vertical: setResponsive('large', 'xlarge') }}
+      role="contentinfo"
+    >
+      <FixedContainer align="center">
+        <Box pad={{ horizontal: setResponsive('0', '0', 'xxxxxlarge') }}>
+          <Row>
+            {/* fixed svalue to preserve UI layout for wider screens */}
```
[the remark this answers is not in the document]
nozomione: Updated, thank you!

src/components/shared/SrOnly.js
nozomione: This is a commonly used name for text specifically intended for screen readers(screen-reader-only), so in this case,`HiddenText` might be too ambiguous. Let's keep this as is or we can also rename it `SROnly`. Let me know, thank you!

src/components/shared/Col.js
nozomione: Previously we talked about this naming and instead of using the short name, we renamed it as [`Column`](https://github.com/AlexsLemonade/refinebio-web/blob/c06a2081f4b498784436e625a09cb07726dee81f/src/components/shared/Column.js) (for PR https://github.com/AlexsLemonade/refinebio-web/pull/32 and up). 

For the flex type, I need to set the [flex-basis](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) and couldn't find the prop for it in the [Grommet Box](https://v2.grommet.io/box#flex). Let me know, thank you!

src/components/shared/Col.js
nozomione: Previously discussed and the wrapper for this component was removed for PR https://github.com/AlexsLemonade/refinebio-web/pull/32 and up (noted [above](https://github.com/AlexsLemonade/refinebio-web/pull/31#issuecomment-1329595464)), thank you!

src/components/shared/Layer.js
nozomione: This component is shared with other components (e.g., [search result's filter](https://github.com/AlexsLemonade/refinebio-web/blob/efe966bfd19d5a9fa2e5925f6cfcdc3b2fdaff23/src/pages/search/index.js)
), and for some cases, the `show` state evaluates whether or not its content is toggled (by default, it's hidden). So fir this reason, simply returning the `children` when `show` is false won't work.

src/components/Header/index.js
```diff
@@ -0,0 +1,46 @@
+import { useState } from 'react'
+import { useResponsive } from 'hooks/useResponsive'
+import { Box, Header as GrommetHeader } from 'grommet'
+import { FixedContainer } from 'components/shared/FixedContainer'
+import styled from 'styled-components'
+import { Logo } from './Logo'
+import { GlobalNav } from './GlobalNav'
+
+const Overlay = styled(Box)`
```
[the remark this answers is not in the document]
nozomione: This `Overlay` was removed (noted [above](https://github.com/AlexsLemonade/refinebio-web/pull/31#issuecomment-1329595464)) in the next PR (when optimizing `Header/*` components), thank you!

src/components/Header/Logo.js
nozomione: Renamed, thank you!

src/components/shared/Layer.js
davidsmejia, line 9:
```suggestion
    if (show) return <GrommetLayer position={position}>{children}</GrommetLayer>
```

This is what I mean

davidsmejia, line 9:
```suggestion
    if (show) return (<GrommetLayer position={position}>{children}</GrommetLayer>)
    return </>
```

If you write it this way you dont have to wrap a single component in a fragment.

src/components/shared/Col.js
davidsmejia: https://v2.grommet.io/box#basis

src/components/shared/Layer.js
davidsmejia: added a suggestion below

davidsmejia (changes requested):
added some clarification, let me know if need anything else

src/components/shared/Layer.js
davidsmejia: This might need to be linted so I wouldn't commit this here

nozomione: Revised, thank you!

**NOTE:**
- Returning the `null` instead of `</>` to avoid having the same eslint error for `jsx-no-useless-fragment`
- The same update was also applied to the latest [`Layer`](https://github.com/AlexsLemonade/refinebio-web/blob/4fe799a11692b236acb47433bbbccd4ef52c784e/src/components/shared/Layer.js) component in PR [37 - Add Responsive to Search Result Page](https://github.com/AlexsLemonade/refinebio-web/pull/38)

nozomione: Resolved in https://github.com/AlexsLemonade/refinebio-web/pull/31#discussion_r1120231495

src/components/shared/Col.js
nozomione: This change will be applied to PR [29 - Homepage (Without Cached Data)](https://github.com/AlexsLemonade/refinebio-web/pull/32) and up, since this component doesn't exist any more, thank you!

src/components/shared/SrOnly.js
nozomione: Removed the wrapper(commented in below), thank you! 
https://github.com/AlexsLemonade/refinebio-web/pull/31#issuecomment-1329595464

davidsmejia (approved):
LGTM

Issue #178, opened 2023-09-13
Implement DatasetContextManager and useDatasetManager


nozomione:
### Context
As with the existing refine.bio, we want to be able to manage users' datasets when they use the application.

### Problem or idea
Using the context API, we'll create a dataset manager context `DatasetManagerContext` and implement its hook `useDatasetManager` which manages a user's dataset and keeps it in sync with the API.

The general strategy should be to have a manager that;

- allow a user to create a dataset and customize it by adding and removing samples, or moving and sharing it with others.
- track a user's dataset ID by storing it in the local storage. Any data associated with the ID must be re-fetched from the API when the user returns (*unless the user clears cache or uses a different web browser/device to access our application).
- keeps the dataset state in sync with the API and queues changes made to an existing dataset

#### <ins>Context State</ins>
The `DatasetManagerContext` inherits the following states from the global context `RefinebioContext` that are shared across the application.

`dataset`: (object) the latest user-created dataset state retrieved from the API 
`setDataset`: an async method for setting `dataset` 

`datasetId`: (uuid) a user-created dataset ID retrieved from the API at creation
`setDatasetId`: an async method for setting `datasetId` 

`downloadOptions`: (object) a user-selected download options for a file download
`setDownloadOptions`: an async method for setting `downloadOptions` 

`email`: (string) a user's provided email address entered in the email form for a file download

`token`: (uuid) a user's token required for a file download

#### <ins>Hook State</ins>
`error`: (object) an object that stores an error object (e.g., network)
`setError`: an async method for setting `error`

`loading`: (boolean) the loading indicator for the dataset data
`setLoading`: an async method for setting `loading`

#### &#10074; Method
The `useDatasetManager` hook contains the helper methods that are primarily used in the UI and each one corresponds to a specific application behavior. 

Following are the implementation details and pseudocode (parameters marked with * are required):

#### Dataset
`async createDataset(setCurrentDatasetId = false) return dataset Id`
- makes API call to create a new dataset
- returns dataset Id

`async getDataset(id) return dataset`
- makes API call to fetch the latest state of the dataset with details
- returns dataset

`async clearDataset(id) return dataset`
- makes API call to clear all the samples in the dataset
- returns dataset

`async updateDataset(id*, params*) return dataset`
- makes API call to updates the dataset with the given dataset Id and params (e.g, sample data, or download options)
- returns dataset

#### Experiment
`getTotalExperiments(data*) return number`
[...]

PR #182, opened 2023-09-14
178 - Implement DatasetManagerContext and useDatasetManager

nozomione:
## Issue Number

Closing #178

The following updates were made in this PR:
- Based on the implementation details, implemented `DatasetManagerContext` and its hook `useDatasetManager`
- Deleted the temporarily added dataset helpers (in `helpers/dataset/*`, total 9 files) and replaced with `useDatasetManager` 
- Cleaned up other temporary code blocks added for demo

## Types of changes

## Checklist


src/contexts/DatasetManagerContext.js
```diff
@@ -1,22 +1,50 @@
...
+  const {
+    dataset: datasetState,
+    setDataset: setDatasetState,
+    datasetId: datasetIdState,
+    downloadOptions: downloadOptionsState,
+    setDownloadOptions: setDownloadOptionsState,
+    setDatasetId: setDatasetIdState,
+    email: emailState,
+    token: tokenState
+  } = useRefinebio()
```
davidsmejia, line 16: I think a better pattern would be to combine all of the global contexts (ex. DatasetManagerContext) into `RefinebioContext` instead of pulling everything from the main context. The benefit of inverting these would mean that `RefinebioContext.Provider` would be the only provider in our  in our `_app.js` and we would still be able to use the hooks as is because they would be scoped correctly.

src/hooks/useDatasetManager.js
```diff
@@ -1,172 +1,216 @@
...
   /* Experiment */
-  const removeExperiment = async (experimentAccessionCode, details = false) => {
-    const data = {}
+  const getTotalExperiments = (data) =>
+    isEmptyObject(data) ? 0 : Object.keys(data).length
+
+  // formats the sample and experiment arrays from the API response
+  // to objects with experiment accession codes as their keys for UI
+  const formatExperiments = (experiments = []) => {
+    if (!experiments.length) return []
```
davidsmejia, line 88: It seems like this function should always return an object? I can't tell if an empty object or the argument `experiments` should be return if it is empty.

Also, I am confused about this function's purpose. It seems like the API's default response is in this format. If this method was called with an empty object it would return an empty array. Below is an example data attribute from a dataset detail view.

```json
  "data": {
    "GSE62528": [
      "GSM1528412",
      "GSM1528409",
      "GSM1528414",
      "GSM1528410",
      "GSM1528413",
      "GSM1528411"
    ]
  },
```

src/contexts/RefinebioContext.js
```diff
@@ -1,20 +1,40 @@
...
 export const RefinebioContext = createContext({})
 
 export const RefinebioContextProvider = ({ children }) => {
+  const [dataset, setDataset] = useLocalStorage('dataset', {})
+  const [datasetId, setDatasetId] = useLocalStorage('datasetId', null)
   const [email, setEmail] = useLocalStorage('user-email', null)
   const [token, setToken] = useLocalStorage('token', null)
+  const [downloadOptions, setDownloadOptions] = useState({})
 
   const value = useMemo(
```
davidsmejia, line 13: See comment in `DatasetManagerContext`

src/components/Header/GlobalNav.js
```diff
@@ -19,25 +18,23 @@ export const GlobalNav = ({ light = false, toggle = false, setToggle }) => {
   const router = useRouter()
   const { asPath, pathname, push } = router
   const { viewport, setResponsive } = useResponsive()
-  // TEMPORARY
-  const { dataset, getDataset } = useDatasetManager()
+  const { dataset, datasetId, getDataset, getTotalSamples } =
+    useDatasetManager()
```
davidsmejia, line 22: Did the formatter/linter do this awkward linebreak?

davidsmejia (changes requested):
Looking really good, just a couple of concerns and requests!

src/contexts/DatasetManagerContext.js
```diff
@@ -1,22 +1,50 @@
...
+  const dataset = datasetState
+  const setDataset = setDatasetState
+  const datasetId = datasetIdState
+  const downloadOptions = downloadOptionsState
+  const setDownloadOptions = setDownloadOptionsState
+  const setDatasetId = setDatasetIdState
+  const email = emailState
+  const token = tokenState
 
   const value = useMemo(
```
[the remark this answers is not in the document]
nozomione: The `dataset` stores the latest state of dataset, thus it's expensive. Let's keep it for this one, but I also agree with your point. Please see the merged [contexts](https://github.com/AlexsLemonade/refinebio-web/tree/dev/src/contexts) in the latest `dev`. All of them currently utilize the `useMemo` hook. 

Let's create a new issue to clean up the  `useMemo` hook from the following contexts 👍:
- `Band Context`
- `ModalContext`
- `SamplesTableManagerContext`
- `SearchManagerContext`

(* also clean up the no longer used `FilterContext` and its hook `useFilter` - replaced with `SearchManagerContext`)

**Side Note:** Our current `eslint` setting enforces the use of `useMemo` for the value object in Context, otherwise the exception comment must be added.

src/contexts/DatasetManagerContext.js
nozomione: The reason I did it this way rather than removing `DatasetManagerContext` is that I wanted to have some flexibility in case the state might only needed to be defined at the `DatasetManagerContext` level, considering it's still in the development stage. Can we keep it for the time being and delete it later as necessary?

src/contexts/RefinebioContext.js
nozomione: My response was added 🗒️

src/components/Header/GlobalNav.js
nozomione: Yes.

src/hooks/useDatasetManager.js
nozomione: Ah, this method is used for the `experiments` array in the API response and not the `data` object (when fetching the dataset with the `details` string query returns the additional properties `experiments` and `organism_samples`).  And an edge case was added so that it returns an empty array if the "experiments" array is empty to prevent an error. 

We use this method to format `experiments` for UI 🐨

e.g. ) 
Before the formatting:
```javascript
[ 
   { …, accession_code: "GSE116436", ... }, 
   { …, accession_code: "SRP066613", ... } 
] 
```

After the formatting:
```javascript
{ 
   GSE116436: {…}, 
   SRP066613: {…} 
} 
```

nozomione: Previously we decided to pass the `details` query to fetch a dataset details view (with `experiments` and `organism_samples`) on every API call and removed the `details` parameter option and `getDatasetDetails` method (https://github.com/AlexsLemonade/refinebio-web/issues/178#issuecomment-1719684364),  but this method was used like so (maybe it was clear as to why we need this):

```js

  // fetches the dataset with the query 'details'
  // details: https://github.com/AlexsLemonade/refinebio-frontend/pull/485
  const getDatasetDetails = async (id = '') => {
    const headers = token
      ? {
          'APT-KEY': token
        }
      : {}
    const response = await api.dataset.get(id, { details: true }, headers)

    const formattedResponse = {
      ...response,
      experiments: formatExperiments(response.experiments) // we reformat the experiments array for UI
    }

    return formattedResponse
  }

```

And due to this update, now this method is used by other methods that make API call to fetch a dataset. For instance, please see `getDataset` method:

https://github.com/AlexsLemonade/refinebio-web/blob/ca24c286f3252a807512e1f7d35ed2a58a9d545f/src/hooks/useDatasetManager.js#L42-L47

src/components/Header/GlobalNav.js
nozomione: This formatting is used (included the code merged into `dev`). Please let me know if it need to be re-configured and if so, let's reformat the files in a new issue.

src/contexts/DatasetManagerContext.js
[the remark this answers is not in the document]
davidsmejia: I will create an issue for this.

src/contexts/DatasetManagerContext.js
davidsmejia: I will create an issue for this.

src/components/Header/GlobalNav.js
davidsmejia: No I think that is fine, just stuck out and was curious.

src/hooks/useDatasetManager.js
davidsmejia: Ok, this seems fine, please have the function return an empty object if it is empty.

davidsmejia:
```suggestion
    if (!experiments.length) return {}
```

davidsmejia (approved):
Please update the `formatExperiments` function to return an empty object and not an empty array (as per suggestion). Then you are free to merge. Otherwise I will open new issues to address other raised concerns.

Issue #179, opened 2023-09-13
Implement useDatasetAction hook for dataset action button


nozomione:
### Context
To make dataset management easy, we provide users three main dataset action buttons in UI which they can use to quickly add or remove downloadable samples to/from their datasets.

The dataset action buttons have three different states, which are as follows:
- "Add to Dataset" (adds all the samples in the selected experiment to a user's dataset)
- "Add Remaining" (displays the number of samples currently added to a user's dataset, and adds the rest)
- "Remove" (removes all the samples in the selected experiment from a user's dataset)

### Problem or idea
We'll implement the hook `useDatasetAction` which contains methods to render the correct button state in UI and to perform its corresponding dataset action. Each method corresponds to a specific application behavior.

#### &#10074; Parameters
Use of this hook requires the following parameters:
`data1`:   a data object in a dataset with `datasetId` stored in a user's localStorage
`data2`:  a data object in the API response selected by a user in UI (e.g., from the search results)

#### &#10074; Method
Following are the implementation details and pseudocode (parameters marked with * are required):

`anyProcessedSamples() return boolean`
returns true if any samples that are processed and available for download in `data2`, otherwise false

`allProcessedInDataset() return boolean`
returns true if all the processed samples in `data2` were added to `data1`, otherwise false

`anyProcessedInDataset() return boolean`
returns true if any processed samples in `data2` were added in `data1`, otherwise false

`getAddedSamples() return samples`
returns the samples in `data2` that were added in `data1`

`getSamplesInDatasets() return samples`
returns all the samples that are in both `data1` and `data2`

`hasSameSamples(d1*, d2*) return boolean`
returns true if two given datasets have the same samples, otherwise false

`intersectDatasets(d1*, d2*) return object`
returns the intersection of two given datasets

`totalSamplesInDataset() return number`
returns the total samples added in `data1`

**Example:** To use this hook, pass the required arguments and destructure the method(s) you want to use:

```js
  const { totalSamplesInDataset } = useDatasetAction(dataset.data, data) 
```

### Solution or next step
Based on the above requirements, implement the hook.

PR #183, opened 2023-09-14
179 - Implement useDatasetAction hook for dataset action button

nozomione:
## Issue Number

Closing #179

Based on the implementation details, implemented `useDatasetAction` hook.

## Types of changes

## Checklist


src/hooks/useDatasetAction.js
davidsmejia, line 1: This looks good!

I would like you to make the following changes to improve readability and make the usage more consistent.

For help understanding which dataset is which, please update the variable / parameter names to:
- `data1` -> `defaultUserDataset`
- `data2` -> `defaultOtherDataset`
- `d1` -> `userDataset`
- `d2` -> `otherDataset`
- derivative variable names should also be updated

davidsmejia (changes requested):
Looking good, please give the variable names in the hook some more meaning and this is good to go!

src/hooks/useDatasetAction.js
```diff
@@ -0,0 +1,105 @@
...
+          data1[experimentAccession],
+          data2[experimentAccession]
+        )
+      }
+    }
+    return data
+  }
+
+  // returns the total samples added in data1
+  const totalSamplesInDataset = () => {
```
davidsmejia, line 94:
```suggestion
  const getTotalSamplesInDataset = () => {
```

Without the get I would assume this is a static value, to make it more obvious that it is a getter please prefix with `get`

src/hooks/useDatasetAction.js
```diff
@@ -0,0 +1,105 @@
...
+  // returns true if any downloadable samples in data2
+  const anyProcessedSamples = () =>
+    Object.values(data2).some(
+      (samples) =>
+        (samples && samples.length > 0) ||
+        (samples && samples.all && samples.total > 0)
+    )
+
+  // returns true if all the processed samples in data2 were added to data1
+  const allProcessedInDataset = () => {
```
davidsmejia, line 17:
```suggestion
  const getAllProcessedInDataset = () => {
```

src/hooks/useDatasetAction.js
```diff
@@ -0,0 +1,105 @@
...
+    if (!anyProcessedSamples()) return false
+
+    const addedSlice = getAddedSamples()
+    return hasSameSamples(addedSlice, data2)
+  }
+
+  // returns true if any processed samples in data2 were added in data1
+  const anyProcessedInDataset = () =>
```
davidsmejia, line 25:
```suggestion
  const getAnyProcessedInDataset = () =>
```

src/hooks/useDatasetAction.js
```diff
@@ -0,0 +1,105 @@
...
+      if (!hasSameElements(d1[accession], d2[accession])) {
+        return false
+      }
+    }
+
+    return true
+  }
+
+  // returns the intersection of two given datasets
+  const intersectDatasets = () => {
```
davidsmejia, line 71:
```suggestion
  const getDatasetIntersection = () => {
```

davidsmejia (changes requested):
While stepping through anther PR I noticed that we should adhere to keeping getters prefixed with `get` so it is clearer that the extracted attribute is a function and not an already computed value.

davidsmejia:
While looking at this. I think that another option besides renaming the functions for clarity would be decide if the value can be derived during execution of the hook, we can just export it as a value and it does not need to be a function. Either that or renaming are acceptable to me.

src/hooks/useDatasetAction.js
nozomione: This hook takes the `data` object in user-created dataset and downloadable samples from the API response (all processed samples in each experiment) which are selected by a user via the search page (no dataset id associated with it but those samples in an experiment were added using the experiment accession code and `all` prop). 

In refinebio-frontend, I see that the term <code>data<strong>Slice</strong></code> is commonly used to distinguish between a `dataset` and a `data` objects in a dataset.

Could we possibly use the `userDataSlice` and `apiDataSlice` (or `dataSliceInExperiment`) or something similar?

nozomione:
@davidsmejia based on your suggestions, I've updated the method names. Please take a look at the comment on parameter names and let me know your thought on this, thank you David.

src/hooks/useDatasetAction.js
davidsmejia: Ah good point, I think instead of `dataslice`, we use `datasetData` as a convention since it is clearer.

Based on your suggestions and my concern I think that `datasetData` and `datasetDataChanges` more clearly illustrate the contents of those variables and also describes how these functions work.

`datasetDataChanges` adequately informs me of the type of data that is expected. That it should look like a datasets `data` attribute and that it is intended to be a potential change to `data` passed as the first parameter.

I can easily see this hook being used anywhere we want to diff changes so lets keep the names distinct from their specific interactions causing their invocation.

davidsmejia (commented):
Name changes are much more descriptive, thanks!

nozomione:
@davidsmejia based on your feedback, I've renamed he parameters for this hook 9144adc542fdcfc7855ea0ace445a5d5c5b56421, please take a look.  

The `hasSameSamples (d1, d2)` is an internal helper method that compares any two `data` objects, so I left the parameter names as they were (If you have any preferences, please let me know).

davidsmejia (commented):
Please update these names to `datasetDataOne` and `datasetDataTwo` in that function and then this looks good to me.

davidsmejia (approved):
LGTM

Issue #187, opened 2023-09-15
Implement the move to dataset modal in UI using useDatasetManager


nozomione:
### Context
We provide users the "Move To Dataset" modal to quickly append or replace their datasets to/with shared datasets (includes processed, unprocessed, and one with processing error) which can be accessed from the dataset page.

Dataset views that renders the move to dataset button:

- Unprocessed shared dataset
- Processed and downloadable dataset
- Processed and expired dataset
- Dataset with processing errors

e.g.) Unprocessed shared dataset view

![Screenshot 2023-09-15 at 4 48 04 PM](https://github.com/AlexsLemonade/refinebio-web/assets/31800566/46dcd12d-4fad-4181-8a15-df33255eb539)
### Problem or idea
We'll need to implement the move to dataset modal functionality utilizing `useDatasetManager` (https://github.com/AlexsLemonade/refinebio-web/issues/178) in the dataset views.

**NOTE:** If there is no user-created dataset, clicking the "Move To Dataset" button will add the shared dataset to my dataset without opening the modal, otherwise the modal will open with options to "append" or "replace" the shared dataset  (matched the current refine.bio behavior).

### Solution or next step
- Utilizing `useDatasetManger`, implement the move to dataset modal functionality
- Create any sub-components or helpers as necessary
- Clean up any temporary code blocks or helpers added for demo

PR #190, opened 2023-09-18
186 - Implement the add page to dataset button in UI using the dataset action button and useDatasetManager

nozomione:
## Issue Number

Closing #186

- This branch was created based on [#189](https://github.com/AlexsLemonade/refinebio-web/pull/189) (will be merged first).
- This PR's commit starts at a4d19bc9327d727fb0a44e30f2677513c01083fd

## Purpose/Implementation Notes
Using the dataset action button and `useDatasetManager`, implemented the add page to dataset button in the search results page.

Please see the latest UI [here](https://refinebio-9hx36ykae-ccdl.vercel.app/).

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

Please attach any screenshots that illustrate these changes.

PR #189, opened 2023-09-15
184 - Implement the dataset action button component in UI using useDatasetManager

nozomione:
## Issue Number

Closing #184

- This branch was created based on [#185](https://github.com/AlexsLemonade/refinebio-web/pull/185) (will be merged first).
- This PR's commit starts at 9a1a0894ea1f2d4ee86ec41bbeecdfd63e93fcae

Main files changed: 6

In `components/shared/DatasetActionButton` (3):
- replaced the temp handle methods in the following components with `useDatasetManger`'s methods:
  - `AddToDatasetButton`
  - `AddRemainingDatasetButton`
  - `RemoveDatasetButton`

In `components/shared/SearchCard` (2):
- `SearchCard.js` (a new file - previously index.js)
   - (minor changes made) destructured the dataset object (for props) and cleaned up the temp code blocks for demo 
- `SearchCardAction.js` (a new file - replaced searchCardCTA.js - created for UI demo)
   - use `DatasetActionButton` and cleaned up the temp code blocks for demo

In `components/shared/SamplesTable` (1):
- `SamplesTableAction` (replaced SamplesTableCTA - created for UI demo)
   - use `DatasetActionButton` and cleaned up the temp code blocks for demo

**NOTE:** The remaining files were simply renamed, deleted (no longer used), or adjusted their import file paths based on that renaming.

## Purpose/Implementation Notes
Based on the requirement, implemented the dataset action button component in UI using `useDatasetManager`.

## Types of changes
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #192, opened 2023-09-18
187 -  Implement the move to dataset modal in UI using useDatasetManager

nozomione:
## Issue Number

Closing #187

- This branch was created based on [#191](https://github.com/AlexsLemonade/refinebio-web/pull/191) (will be merged first).
- This PR's commit starts at 789c55f37db0015ceec43adf95dbfdf716c34e0a

Using `useDatasetManager`, implemented the move to dataset button functionality for the shared datasets.

Please see the latest UI [here](https://refinebio-ah95asro7-ccdl.vercel.app/).

To test this button, please use the below dataset IDs or use any datasets you've created (a URL path should be `VERCEL_PREVIEW_URL/dataset/DATAET_ID`):

Shared dataset id `b62c4aa6-2332-4d1a-ad67-4be26909efdd` with 6,633 samples:
https://refinebio-ah95asro7-ccdl.vercel.app/dataset/b62c4aa6-2332-4d1a-ad67-4be26909efdd?ref=share

Shared dataset `ff9f34ad-b69d-403b-a9f9-07f0b5739b7a` with 165 samples:
https://refinebio-ah95asro7-ccdl.vercel.app/dataset/ff9f34ad-b69d-403b-a9f9-07f0b5739b7a

**NOTE:** If there is no user-created dataset, clicking the "Move To Dataset" button will add the shared dataset to my dataset without opening the modal, otherwise the modal will open with options to "append" or "replace" the shared dataset (matched the current refine.bio behavior).

## Types of changes

## Checklist


src/components/SearchResults/AddPageToDatasetButton.js
```diff
@@ -0,0 +1,28 @@
...
+  const { setResponsive } = useResponsive()
+
+  const downloadableSamples = fromPairs(
+    dataToAdd
+      .filter((item) => item.num_downloadable_samples > 0)
+      .map((item) => [
+        item.accession_code,
+        { all: true, total: item.num_downloadable_samples }
+      ])
+  )
```
davidsmejia, line 15: We should move the logic of creating this data structure to a method in `useDatasetAction.js`. Since it is only consumed there it will help understanding how those functions get that unique data structure.

This will mean an additional change in the hook to not to expect "preformatted" ie not expect `all` and `total` to be attributes on the passed in array.

I would expect something like this:
```suggestion
  const downloadableSamples = getFormattedDatasetData(
    dataToAdd
      .filter((item) => item.num_downloadable_samples > 0)
  )
```

Then in `useDatasetActions`
```js
const getFormattedDatasetData = (experimentList) =>
  Object.fromEntries(
    experimentList.map((item) => [
      item.accession_code,
      { all: true, total: item.num_downloadable_samples }
    ])
  )
```

src/components/shared/SamplesTable/SamplesTableAction.js
```diff
@@ -19,6 +19,7 @@ export const SamplesTableAction = ({ accessionCode, downloadableSamples }) => {
         data={{
           [accessionCode]: { all: true, total: downloadableSamples }
```
davidsmejia, line 20: The same thing applies here as above. Lets use the same function as described above to generate this structure.

src/components/Dataset/MoveToDatasetModal.js
```diff
@@ -10,49 +10,53 @@ import { Row } from 'components/shared/Row'
...
-      router.push(
+      push(
         {
-          pathname: '/download',
+          pathname,
           query: {
-            message: `Appended ${formatNumbers(
-              getTotalSamples(dataset.data)
-            )} samples to My Dataset`,
+            message: `Appended ${newDatasetTotalSamples} samples to My Dataset`,
```
davidsmejia, line 40: We should avoid passing in formatted strings as query parameters. We should instead pass the data for the component to use as needed.
 
 This applies to line 52 below as well. 

```suggestion
            samplesCount: newDatasetTotalSamples,
            action: "Appended",
```

davidsmejia (changes requested):
Looking good, I have a couple change requests, let me know if you have any questions.

src/pages/dataset/[dataset_id].js
```diff
@@ -17,86 +19,127 @@ import {
...
-  const [data, setData] = useState(null)
+  const [selectedDataset, setSelectedDataset] = useState({})
 
   useEffect(() => {
-    if (pageRendered) {
-      setData(dataset)
+    const getSelectedDataset = async (id) => {
+      const response = await getDataset(id)
+      setSelectedDataset(response)
+      return response
```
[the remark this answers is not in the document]
Comment (nozomione) [reply src/pages/dataset/[dataset_id].js:49]: Yes, I returned it just in case I needed for a later implementation. I'll make sure to remove it once the dataset page implementation is completed 👍

src/components/SearchResults/AddPageToDatasetButton.js
nozomione: @davidsmejia, these implementations were approved in previous PRs https://github.com/AlexsLemonade/refinebio-web/pull/189 and https://github.com/AlexsLemonade/refinebio-web/pull/190 3 days ago and already merged into `dev`. 

Since this PR is for the move to dataset modal implementation, can we tackle this in a separate issue to keep the context as specific as possible? Please let me know, thank you David.

src/components/shared/SamplesTable/SamplesTableAction.js
nozomione: Please see my comment above, thank you David.

src/components/SearchResults/AddPageToDatasetButton.js
nozomione: I can create a new issue to update this tomorrow?

src/components/Dataset/MoveToDatasetModal.js
nozomione: The parameter `message` (its value as template literals) is used in the shared [`Notification`](https://github.com/AlexsLemonade/refinebio-web/blob/dev/src/components/shared/Notification.js) component which allows this shared component to be flexible and generic (takes any messages, not specific data values as parameters).

The router's `query` object in our application is "watched" by `Notification` from `Layout` (at the top level). When the `message` and `status` are added to the `query` object, that `message` value will be rendered in UI using the matching theme for the `status` (for example, `success` is green, `error` is coral). 

If we only pass the data value instead of the whole message value, we would have to build a new custom component in order to render the exact message that goes with that data value. 

For instance, for ths move to dataset modal, we'll need to pass the data value for `newSampleCounts`, `action`, and `status` etc, and if we need to display another notification for a particular feature, each of these also needs its own special custom component and specific parameters (and that also makes our codebase less DRY).

Could we please keep this `message` parameter? Or please let me know if there is a better alternative 💡

davidsmejia: I see what you mean.

Placing that value in local storage via a context would be more consistent with how we manage session variables.

I think this is ok for now and moving that over can be done in another ticket.

We need to add a default option for `status` in the `Notification` component or null check because it crashes if only `message` is in the query params as it is right now.

Please add that in this PR and then this is good for now.

src/components/SearchResults/AddPageToDatasetButton.js
davidsmejia: Yeah can you apply my request to those as well in this PR after that this looks good to me.

src/pages/dataset/[dataset_id].js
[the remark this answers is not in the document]
Comment (davidsmejia) [reply src/pages/dataset/[dataset_id].js:49]: Yeah more of a housekeeping request. Doesn't have to be in this PR so please if there is one that is open that would be better please address it there or leave a comment that where it should be done.

davidsmejia (changes requested):
Would like to see the logic for creating those data structures in `useDatasetAction` otherwise looks good to me!

nozomione:
> Would like to see the logic for creating those data structures in `useDatasetAction` otherwise looks good to me!

Great, I'll revise the `useDatasetAction` in this new issue https://github.com/AlexsLemonade/refinebio-web/issues/210 (with a new feature branch to make the merge process a bit simpler), thank you David!

nozomione:
🗒️ We should improve this implementation and create a notification manager which manages the application notifications and is responsible for toggling notification popups.

davidsmejia (approved):

Issue #181, opened 2023-09-13
Implement add and remove sample buttons in Samples Table UI using the dataset action button component and useDatasetManager


nozomione:
### Context
As with the existing refine.bio, we want to be able to let users quickly add or remove a single downloadable sample from the samples table. 

**NOTE:** The the add/remove buttons in the samples table are accessible from the experiment and download page. However, the dataset page's the samples table data is immutable. 

Add and remove a sample buttons in the samples table UI:

![Screenshot 2023-09-13 at 6 52 48 PM](https://github.com/nozomione/test/assets/31800566/5cd86f49-f745-4d57-a38b-f4d34e832803)

### Problem or idea
Using the dataset action button component(https://github.com/AlexsLemonade/refinebio-web/issues/180) and `useDatasetManager`(https://github.com/AlexsLemonade/refinebio-web/issues/178), we'll implement the add and remove a sample buttons in the samples table.

🗒️  This implementation needs the API cache to perform the check to see if the samples have qn targets associated, however we'll use the mock data for it until PR https://github.com/AlexsLemonade/refinebio-web/pull/164 is merged in `dev`. 

### Solution or next step
- Utilize `useDatasetManager` to implement the add and remove a sample button component in the sample stable
- Create any sub-components or helpers as necessary
- Clean up any temporary code blocks or helpers added for demo

PR #194, opened 2023-09-19
181 -   Implement add and remove sample buttons in Samples Table UI using the dataset action button component and useDatasetManager

nozomione:
## Issue Number

Closing #181

- This branch was created based on [#192](https://github.com/AlexsLemonade/refinebio-web/pull/192) (will be merged first).
- This PR's commit starts at 38cfeeba8bcc517266c8b21d15bbd4d4f9c0212f

Implemented the add/remove button functionality for the samples table in the experiment and download pages.

Please see the latest UI [here](https://refinebio-gmybjdelv-ccdl.vercel.app/).

**NOTE:** This implementation requires the API cache to perform the check to see if the samples have qn targets associated, however currently the mock data is used until PR https://github.com/AlexsLemonade/refinebio-web/pull/164 is merged in `dev`. 

## Types of changes

## Checklist


davidsmejia (approved):
This looks good.

I am going to note here that the references to the cache will change because we should not be updating this value only on client side deploy. For now this is ok if you prefer to have it this way fist. I would have preferred to implement it by fetching the api response on every load, then adding the cache in later. If you can make that change please do otherwise we can rework it at a later time

Issue #207, opened 2023-10-09
Implement the request data form in search results page


nozomione:
### Context
By submitting the request data form, users can request data that are not yet available in our application. The form is accessible when users' search yields no results on the search results page.

### Problem or idea

Data request form:

![Screenshot 2023-10-09 at 11 10 28 AM](https://github.com/AlexsLemonade/refinebio-web/assets/31800566/f91bdca9-bfb3-4e08-b778-caf7d791c49b)

### Solution or next step
- Implement the request data form when no search results
- Add form validation utilizing [formki](https://www.npmjs.com/package/formik) and [yup](https://github.com/jquense/yup)
- Create any sub-components or helpers as necessary
- Clean up any mock data or temp code blocks created for demo

PR #226, opened 2023-10-25
207 - Implement the request data form in search results page

nozomione:
## Issue Number

Closing #207

**NOTE:** The **changes in 21 files** were simply the result of file deletions, file renaming, import adjustments, and minor style tweaks.

- Implemented the request data form for no search results in search results page
- Added the form validation schema to the request data form 

Please see the latest UI [here](https://refinebio-6ux14uh7h-ccdl.vercel.app/).

To display the request data form using an accession code, please use the accession code `SRP2422`: 
https://refinebio-6ux14uh7h-ccdl.vercel.app/search?search=SRP2422

To display the request data form using a search keyword, please use a term like `xyz`:
https://refinebio-6ux14uh7h-ccdl.vercel.app/search?search=xyz

To prevent spams, the following checks were added (chatted with Deepa):
- Added a validation check for the experiment accession codes text field to prevent users from entering random characters  (current refine.bio includes an empty value check, but no accession code format)
- Set the character limit to 255 for the open-ended text field ("Is there anything else you would like to add?")

The following fixes are included in this PR to properly return search results and render the data request form when there are no results:
- **Bug Fix 1**  toggling each search filter by chars (293d5fd20931ffb0752ed28a098eea4689b9ce58)
- **Bug Fix 2**  fetching accession codes on search results (b93d064a52f8df92a2a93d352f5fd19f8cc35add)
- **Bug Fix 3**  fetching the array express accession code (301489ef84456d3ea09317ec24141208ae24bccb)

<details> 

<summary><strong>More Context for Bug Fix 3:</strong></summary>

Currently refine.bio doesn't return the search result for [the experiment for `E-AFMX-1`](https://www.refine.bio/experiments/E-AFMX-1/transcription-profiling-of-human-chimp-and-mouse-brain) when searched via the search box even though it's available via [API](https://api.refine.bio/v1/search/?search=accession_code:E-AFMX-1&search=alternate_accession_code:E-AFMX-1).

Screenshot 2023-10-26 at 9 12 03 AM

However, we can use the facets to find that experiment (applied filters: `Includes Publication` > `Macaca mulatta`):

Screenshot 2023-10-26 at 8 46 15 AM

The reason was that FE's [regex for the experiment accession code](https://github.com/AlexsLemonade/refinebio-frontend/blob/488101b32df40157729e2fb24e3bfa4aa382d069/src/state/search/actions.js#L88) doesn't include a single digit case for the array express accession code, so the application doesn't make the API request for fetching accession codes (*FE makes multiple API calls for fetching accession codes). For this reason, to make sure our application makes the correct API calls, I've updated the regex to include a single digit case.

Please see the search results for `E-AFMX-1` on Vercel preview [here](https://refinebio-6ux14uh7h-ccdl.vercel.app/search?search=E-AFMX-1):

Screenshot 2023-10-26 at 9 00 44 AM
<hr />
</details>

Other minor adjustments:
- Fix the bug on downloadable samples count (74781a8b125548136f27a66080337d3a388fa797)
- Replace the array express accession code URL with the correct one (for alternate accession IDs, e.g., 'E-') (f0ad19f0881fb76c1f6f5226e138f4ec668e4788) (credit : Deepa)

## Types of changes

## Checklist


davidsmejia (commented):
Could you explain why we are handling accession code matches in fetch search in a PR about the request data form?

nozomione:
> Could you explain why we are handling accession code matches in fetch search in a PR about the request data form?

I've added the accession code bug fixes to this PR since the data request form for search results depends on the accuracy of the search results. Please let me know if you have question regarding the the explanation I've added in the implementation note section.

nozomione:
@davidsmejia  I've just made a small tweak to add a current page check for accession codes API requests (please see the last commit 5f9d1daca242b918ccbc16a72ffaf3938734c6cc). I was checking the search on refine.bio and realized that this check was necessary, thank you.

src/components/SearchResults/SearchFilterList/SearchFilter.js
```diff
@@ -26,54 +26,77 @@ const ToggleButton = styled(sharedButton)`
...
+            (filterOption === 'platform'
+              ? formatPlatformName(cache.platforms[option[0]]) || option[0]
+              : formatString(option[0])
+            )
+              .toLowerCase()
+              .includes(val.toLowerCase())
+          )
+        : open
+        ? filterList
+        : filterList.slice(0, maxCount)
```
davidsmejia, line 54: This is a bit much. Can we break this up to determine what to set instead of having all of the logic in this function?

src/helpers/fetchSearch.js
```diff
@@ -1,24 +1,43 @@
...
+    if (matchedAccessionCodes.length > 0) {
+      // removes duplicates due to multiple requests for fetching accession codes
+      results = getUniqElementsBy(
+        [...matchedAccessionCodes, ...results],
+        (x) => x.accession_code
+      )
+    } else if (accessionCodes.length === 1) {
+      // returns no result if no match found for a single accession code that users searched for
+      results = []
+    }
```
davidsmejia, line 39: I'm going to open a ticket because this approach is ill-advised for the following reasons.

Given that:
- N is the number of accessions parsed from the query.
- X is the number of results per search result page.

1. This prepends the matched accession codes to the page of the results. That page states to the user that we are showing X results per page. With this implementation we will be showing X + N results on the page where N is the number displayed on the page.
1. This approach is unnecessarily expensive as it will create N + 1 queries per result when this could be 1.

I think this could be resolved with 2 changes.
1. Construct a query that uses the built in `accession_code__in` and `alternate_accession_code__in` query parameters. To fetch all experiments specifically mentioned in the query string.
1. Add some visual distinction between the augmented matches at the top and the actual search results. This would be a Deepa question.

src/pages/search/index.js
```diff
@@ -170,40 +168,39 @@ export const Search = (props) => {
...
-                )}
                 <Box animation={{ type: 'fadeIn', duration: 300 }}>
-                  {results.results.map((result) => (
-                    <SearchCard key={result.id} result={result} />
-                  ))}
-                  {results.results.length < 10 && <MissingResultsAlert />}
+                  {results.map((result, i) =>
+                    result.isMatchedAccessionCode ? (
+                      <Fragment key={result.id}>
+                        <SearchCard key={result.id} result={result} />
```
davidsmejia, line 182: Broadly speaking we should be using `result.accession_code` as the unique ID. Because this value isn't guaranteed to be persistent between reindexing. It should be removed from the API search response in the future if possible.

This probably isn't going to be a problem here today. However, consistency in using `accession_code` would be preferred as it could lead to issues if used elsewhere.

davidsmejia (changes requested):
Please fix the spelling of `RequestSearchFormAlert` and make the size `16px`.  Then re-request review. Otherwise the other comments can be addressed in the future.

Issue #250, opened 2023-11-03
Fix: Correct the text color of the non-linked info massage


nozomione:
### Context

All the non-linked info messages's color needs to be adjusted from the brand color (blue) to black. 

### Problem or idea
We should fix the non-linked info message's color to black (body font color).

e.g., )
The dataset action buttons:
<img width="570" alt="Screenshot 2023-11-03 at 12 12 16 PM" src="https://github.com/AlexsLemonade/refinebio-web/assets/31800566/7e8c1681-25cc-4ddb-bc44-882976202c70">

The samples table:
<img width="648" alt="Screenshot 2023-11-03 at 12 12 02 PM" src="https://github.com/AlexsLemonade/refinebio-web/assets/31800566/9274df87-26a2-4f3d-8647-77eb661e5ad5">

The additional metadata modal: 
<img width="1321" alt="Screenshot 2023-11-03 at 12 08 15 PM" src="https://github.com/AlexsLemonade/refinebio-web/assets/31800566/5ed33d3c-b9cc-4ec0-a8ff-995371393c3a">

### Solution or next step
- Specify the color to black for the non-linked info messages

PR #257, opened 2023-11-06
250 - Fix: Correct the text color of the non-linked info massage

nozomione:
## Issue Number

Closing #250

Corrected the text color of the non-linked info massage.

## Types of changes
-Bugfix (non-breaking change which fixes an issue)

## Checklist


src/components/shared/InlineMessage.js
```diff
@@ -33,7 +33,7 @@ export const InlineMessage = ({
           <Icon color={color} name={SVG} size={iconSize} />
         </Box>
       )}
-      <Paragraph color={color || fontColor} size={fontSize}>
+      <Paragraph color={color !== 'info' ? color : fontColor} size={fontSize}>
```
davidsmejia, line 36: If I were to set the `fontColor` but not `color`, then there would be no effect. I think what we should do here is just set the default `fontColor` to info an explicitly set it when we want it to change.

davidsmejia: Correction: if we set color and fontColor, fontColor would have no effect

nozomione: I've defined the `types` to specify default settings for each `type` and adjusted the codebase and is now ready for another look. Good call! Thank you @davidsmejia 👍

davidsmejia (approved):
LGTM

PR #272, opened 2023-11-09
271 - Implement error handling for API resource requesets for the search results page

nozomione:
## Issue Number

**NOTE:** The **changes in n files** were simply the result of file/variable name renaming, import adjustments, and minor copy updates.

Closing #271

Implement error handling for the search results page and render an error view based on the status code when error occurred.

Please see the latest UI [here]().

## Types of changes

## Checklist


davidsmejia (commented):
I think this looks good to me. My only concern is around the renaming of `API_HOST` and `API_PATH`.

Existing convention:
- `API_HOST`: The domain of our API, localhost, staging.refine.bio, refine.bio
- `API_VERSION`: The path partial of the API where API endpoints are nested.
- `API_PATH`: Root path of all API endpoints, this is the concatenated host and version.

I think removing the protocol from `API_HOST` would be something to be considered but I do not understand the intent behind this change in this PR.

nozomione:
> I think this looks good to me. My only concern is around the renaming of `API_HOST` and `API_PATH`.
> 
> Existing convention:
> 
>     * `API_HOST`: The domain of our API, localhost, staging.refine.bio, refine.bio
> 
>     * `API_VERSION`: The path partial of the API where API endpoints are nested.
> 
>     * `API_PATH`: Root path of all API endpoints, this is the concatenated host and version.
> 

I am sorry for the typo. I just fixed the value of  the `apiPath` variable in `sitemap.config` (9c1694ceade37bdbff2e691cca2f89c461fd22b0) and this was the reason for renaming env name (to `API_HOST` from `API_PATH`). 

> I think removing the protocol from `API_HOST` would be something to be considered but I do not understand the intent behind this change in this PR.

I cleaned up the declaration of `STAGE_API_PATH` and `STAGE_API_VERSION` variables from the `docker-compose.env` file (051589656bc0c69b881fc3da88b89b9370075f9b) since currently we are not using them to compose docker container.  But I kept the protocol (`https`) and was added in all env variables in Vercel. 

Also I ran the deploy scripts locally since the project counts changed for sitemap etc for Vercel preview deploys (02d87b9f92cdf0ff15b48ff39c95d90cb94a149b). Thank you!

davidsmejia (approved):

Issue #279, opened 2023-11-13
QA - FIlter: Rename Rna-seq to RNA-seq


dvenprasad:
### Context

Filter value for RNA-seq is formatted incorrectly.

![Screenshot 2023-11-13 at 09 19 58](https://github.com/AlexsLemonade/refinebio-web/assets/15315514/ab26fdf4-c957-455f-ab1e-d01332671e61)

### Problem or idea

It should be `RNA-seq` instead.

### Solution or next step

Tagging @davidsmejia for next steps

PR #283, opened 2023-11-13
279 - QA: FIlter Rename Rna-seq to RNA-seq

nozomione:
## Issue Number

Closing #279

Rename the `Rna-seq`name  to be `RNA-seq` in UI.

@dvenprasad Please see the latest UI [here](https://refinebio-b1uzwah0s-ccdl.vercel.app/).

## Types of changes

What types of changes does your code introduce?

## Functional tests

List out the functional tests you've completed to verify your changes work locally.

## Checklist


dvenprasad (approved):
👍 

🐮 🐮 🐮

davidsmejia (commented):
I'd like to see some changes before we merge this:

- move the `getFormattedTechnologyNames` to `helpers/formatTechnologyName`
- remove the nested ternary operator here and handle this by either:
  - test for default case first ie. `!['platform', 'technology'].includes(filterOption) and then short circuit for the specific special cases
  - alternatively, move this logic to a function at the top of the component that takes `filterOption` and `option`

Additionally, it looks like `helpers/formatPlatformName` correctly falls back to the passed in value so we can remove the `|| option[0]` fall back.

Looking at `helpers/formatPlatformName`, you could further clean it up by setting the default value to an empty string and removing that first check but that change doesnt need to occur in this PR.

nozomione:
> I'd like to see some changes before we merge this:
> 
>     * move the `getFormattedTechnologyNames` to `helpers/formatTechnologyName`
> 
>     * remove the nested ternary operator here and handle this by either:
>       
>       * test for default case first ie. `!['platform', 'technology'].includes(filterOption) and then short circuit for the specific special cases
>       * alternatively, move this logic to a function at the top of the component that takes `filterOption` and `option`
> 
> 
> Additionally, it looks like `helpers/formatPlatformName` correctly falls back to the passed in value so we can remove the `|| option[0]` fall back.
> 
> Looking at `helpers/formatPlatformName`, you could further clean it up by setting the default value to an empty string and removing that first check but that change doesnt need to occur in this PR.

Thank you for the idea💡! 

I made the following adjustments:
- moved the `getFormattedTechnologyNames` to `helpers/formatTechnologyName`
- added the new `helpers/formatFilterNames` which is responsible for formatting all the filter names for UI
   - defined the constants in `config/options` for the filter option names that need to be formatted 
- used the`helpers/formatFilterNames` in the `SearchFilter` component and cleaned up the ternary operators for filter names

Ready for another look, thank you!

src/helpers/formatFilterNames.js
```diff
@@ -0,0 +1,16 @@
+import { options } from 'config'
```
davidsmejia, line 1:
```suggestion
```

src/components/SearchResults/SearchFilterList/SearchFilter.js
```diff
@@ -95,10 +95,11 @@ export const SearchFilter = ({
...
                     <TextHighlight>
-                      {filterOption === 'platform'
-                        ? formatPlatformName(cache.platforms[option[0]]) ||
-                          option[0]
-                        : formatString(option[0])}
+                      {formatFilterNames(
+                        filterOption,
+                        option[0],
+                        cache.platforms[option[0]]
+                      )}
```
davidsmejia, line 102:
```suggestion
                      {formatFilterName(filterOption, option[0])}
```

src/components/SearchResults/SearchFilterList/SearchFilter.js
```diff
@@ -4,9 +4,8 @@ import styled, { css } from 'styled-components'
 import { useResponsive } from 'hooks/useResponsive'
 import { useSearchManager } from 'hooks/useSearchManager'
 import { TextHighlightContextProvider } from 'contexts/TextHighlightContext'
+import formatFilterNames from 'helpers/formatFilterNames'
```
davidsmejia, line 7:
```suggestion
import formatFilterName from 'helpers/formatFilterName'
```

src/components/SearchResults/SearchFilterList/SearchFilter.js
```diff
@@ -47,9 +46,10 @@ export const SearchFilter = ({
...
       val.trim() !== ''
         ? filterList.filter((option) =>
-            (filterOption === 'platform'
-              ? formatPlatformName(cache.platforms[option[0]]) || option[0]
-              : formatString(option[0])
+            formatFilterNames(
+              filterOption,
+              option[0],
+              cache.platforms[option[0]]
             )
```
davidsmejia, line 53:
```suggestion
            formatFilterName(
              filterOption,
              option[0]
            )
```

src/helpers/formatTechnologyName.js
```diff
@@ -0,0 +1,7 @@
+import formatString from 'helpers/formatString'
+
+export default (option) => {
+  const rnaSeq = 'RNA-seq' // 'rna-seq' should be rendered as 'RNA-seq' in UI
+
+  return option.match(rnaSeq.toLowerCase()) ? rnaSeq : formatString(option)
+}
```
davidsmejia, line 7: `technologyNames` would make sense to put in config now

```suggestion
const technologyNames = {
  'rna-seq': 'RNA-seq'
}
export default (option) => {
  return technologyNames[option.toLowerCase()] || formatString(option)
}
```

src/helpers/formatFilterNames.js
```diff
@@ -0,0 +1,16 @@
...
+  const { filterOptionsToFormat } = options.search
+  const formatName = () =>
+    filterOption === 'platforom'
+      ? formatPlatformName(formattedPlatformName)
+      : formatTechnologyName(item)
+
+  return !filterOptionsToFormat[filterOption]
+    ? formatString(item)
+    : formatName()
+}
```
davidsmejia, line 16: I think we can make this easier to understand. This function really just needs to take a `filterOption` and map that to a corresponding formatting function. 

This means that we will need to move the cache.platforms logic to the `formatPlatformName` helper. That function should still just take `platformName` as the only argument only now it will check to see if it is in the cache then handle the value the same as it currently does.

```suggestion
export default (filterOption, name) => {
  
  if (filterOption === 'platform') {
    return formatPlatformName(name)
  }
  
  if (filterOption === 'technology') {
    return formatTechnologyName(name)  
  }

  return formatString(item)
}
```

davidsmejia (commented):
Can we make the helper to `formatFilterName` since it only handles a single name.

davidsmejia (approved):
LGTM

src/helpers/formatFilterNames.js
nozomione: Now using the API cache directly in `helpers/formatPlatformName` and refactored `helpers/formatFilterName` with the switch statement for readability and maintainability. Thank you for the feedback!

Issue #292, opened 2023-11-16
Fix: Implement the dataset loader for polling the latest dataset state for the dataset page UI


nozomione:
### Context
When users request to start processing a dataset for download, the application displays the dataset processing view (issue: https://github.com/AlexsLemonade/refinebio-web/issues/200). When the dataset has been processed, it displays either the ready for download view (issue: https://github.com/AlexsLemonade/refinebio-web/issues/201) or the processing error view (https://github.com/AlexsLemonade/refinebio-web/issues/44), depending on that processing result.

Currently, this feature doesn't work in [refine.bio](https://www.refine.bio/) as well as in this `refinebio-web` repo and it needs to be fixed.

### Problem or idea
We should implement a logic to poll the dataset processing status and render an appropriate view in the dataset page.

### Solution or next step
(*due to frequent PR merges, the most recent PR's feature branch will be used as the base branch to reduce merge conflicts)

- Create or reuse a component for the dataset loader
- Create any helpers as necessary

PR #215, opened 2023-10-13
166 - Implement useOneOffExperiment hook for one-off experiment dataset downloads

nozomione:
## Issue Number

Closing #166

## Purpose/Implementation Notes
Implemented the hook `useOneOffExperiment` which manages the file download for one-off experiment dataset. 

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

**NOTE:** The`Notification` and `[dataset_id]` files are included in the files changed count but they can be safely ignored. They will be removed after the previous base branch is merged into `dev` (this previous base branch hasn't  yet pulled the most recent `dev` changes).

Commits in this PR:
- af9193270cb08ee1eb8bf0914b6f41de626eae04
- 43b0c47ce639d7fb3f415fefacb4ed181818194c
- 6d8a3bbd01cdccae2fc5b00b068b44ac6a72f5aa
- aeb7e7b4dc97f999179190d132cc9edbd8660409

Issue #287, opened 2023-11-14
Fix: Adjust the regenerate dataset feature and UI data rendering 


nozomione:
### Context

Related issue: #281

Currently, the regenerate dataset feature is broken in [refine.bio](https://www.refine.bio/dataset/dfaf1b2f-6d13-4695-8eb3-b1b8931c7210), as well as in this `refinebio-web` repo (due to the recent implementation change in #286, the logic must be adjusted) and it needs to be fixed.

![Screenshot 2023-11-14 at 4 32 50 PM](https://github.com/AlexsLemonade/refinebio-web/assets/31800566/f5a93372-c906-43c8-a098-04edfdede4ca)

### Problem or idea
Implement the regenerate dataset feature and render the correct dataset state in UI per download options change.

### Solution or next step
- Implement the regenerate dataset feature 
- Create any sub-components or helpers as necessary

PR #294, opened 2023-11-21
292 - Fix: Implement the dataset loader for polling the latest dataset state for the dataset page UI

nozomione:
## Issue Number
**NOTE:** The **changes in 5 files** were simply the results of fixing hook imports and argument orders, deletion of a file, and state imports. 

Closing #292

Implemented logic for polling the latest state of the processing dataset, and based on that state, render the appropriate view (processing/ready for download/error) on the dataset page.

- Created the `useResourceLoader` hook (refactored [`useOneoffExperiment`](https://github.com/AlexsLemonade/refinebio-web/pull/215) hook) for downloading files for both user-created datasets and one-off experiments datasets. 

Please see the latest UI [here](https://refinebio-clz6yymct-ccdl.vercel.app/).

For the user-defined datasets (My Dataset):
**Step 1:** Add samples to My Dataset 
**Step 2:** Start the dataset download process by clicking the "Download" button in the dataset page
**Step 3:** Once the dataset processing is completed, the application renders the appropriate UI (success/error)

Screenshot 2023-11-20 at 9 17 59 PM

<br/><br/>
For the one-off experiment datasets:
**Step 1:** Click the "Download Now" button in a search card on the search results page to start the download process
**Step 2:** The application renders the processing budge while the one-off experiment dataset being processed

Screenshot 2023-11-21 at 2 01 37 PM

(search results page: while processing)

image

(experiment page: while processing)

image

## Types of changes

## Functional tests

Tested the implementation and made sure it works for both My Dataset download and the dataset download for one-off experiment. 

## Checklist


src/components/Dataset/DatasetPageHeader.js
```diff
@@ -32,8 +32,7 @@ const Block = ({ children }) => {
   )
 }
 
-export const DatasetPageHeader = ({ dataset }) => {
-  const { error } = useDatasetManager()
+export const DatasetPageHeader = ({ dataset, hasError }) => {
```
davidsmejia, line 35: I'm trying to understand why we must pass in `hasError` and wouldn't be able to derive it from the `dataset` prop.

nozomione: The `useResourceLoader` hook is used by different components (including this `DatasetPageHeader`, [DownloadNowModal](https://github.com/AlexsLemonade/refinebio-web/blob/b45ba3d342e247f094c3ece4bcb79bc509e5a437/src/components/shared/SearchCard/SearchCardAction/DownloadNowModal.js#L24-L26), and the [experiment](https://github.com/AlexsLemonade/refinebio-web/blob/b45ba3d342e247f094c3ece4bcb79bc509e5a437/src/components/shared/SearchCard/SearchCardAction/DownloadNowModal.js#L24-L27) page components) to "watch" the latest processing status of resources (a user-custom dataset and an one-of experiment download's dataset). The `hasError` state is set to true if API requests fail while polling the data, and it could be used to display a error message in UI etc.  

In terms of the `dataset` state, it's a part of `useDatasetManager` and is used to render the dataset's data in UI.

This implementation (utilizing the `hasError` state) also matches other hooks' pattern ([`useComendia`](https://github.com/AlexsLemonade/refinebio-web/blob/118c281d226a70d6cdc972c290b87b25c93c6b18/src/hooks/useCompendia.js#L20), [`useLoader`](https://github.com/AlexsLemonade/refinebio-web/blob/118c281d226a70d6cdc972c290b87b25c93c6b18/src/hooks/useLoader.js#L10), and [`useSamplesTable`](https://github.com/AlexsLemonade/refinebio-web/blob/118c281d226a70d6cdc972c290b87b25c93c6b18/src/hooks/useSamplesTableManager.js#L17), and the helper [`fetchSearch`](https://github.com/AlexsLemonade/refinebio-web/blob/118c281d226a70d6cdc972c290b87b25c93c6b18/src/helpers/fetchSearch.js#L62) for search results page that fetches API data via its `getInitialProps`). 

For this, I used the `useResourceLoader` hook's `hasError` (instead of `useDatasetManager`'s `error` object - please see the PR https://github.com/AlexsLemonade/refinebio-web/pull/297) and pass it to `DatasetPageHeader` (which is responsible for re-rendering UI based on the most recent dataset status that was last polled from API) via the [`dataset_id`](https://github.com/AlexsLemonade/refinebio-web/blob/b45ba3d342e247f094c3ece4bcb79bc509e5a437/src/pages/dataset/%5Bdataset_id%5D.js#L33-L38) page component.

davidsmejia (commented):
Left some initial thoughts here. They mostly focus on the `useResourceLoader` hook.

src/components/Dataset/DatasetPageHeader.js
```diff
@@ -46,46 +45,36 @@ export const DatasetPageHeader = ({ dataset }) => {
...
+            </Block>
+          ))
+        ) : (
+          dataset?.data && (
+            <Box pad={{ top: 'large', bottom: 'medium' }}>
+              <Heading level={2} size={setResponsive('small', 'large')}>
+                Shared Dataset
+              </Heading>
+            </Box>
+          )
```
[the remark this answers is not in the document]
nozomione: I've reverted the JSX changes (at 706548df0f631f4b63a7708d2882de9168563564) and it now uses early returns 👍

src/hooks/useResourceLoader.js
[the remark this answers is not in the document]
nozomione: Renamed af3b4c72d78188d80344fbfa73f991f148109e9b.

src/hooks/useResourceLoader.js
```diff
@@ -0,0 +1,80 @@
...
+    setProcessingResources((prev) => {
+      if (prev.find((item) => item.datasetId === datasetId)) return prev
+
+      return [...processingResources, { datasetId, accesionCode }]
+    })
+  }
+
+  // id: a dataset ID || an experiment accession code
+  const getProcessingResource = (id) => {
+    const keyToFind = oneOffExperiment ? 'accesionCode' : 'datasetId'
```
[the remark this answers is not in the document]
nozomione: I've remove this optional parameter `oneOffExperiment` and instead, now using the areValidAccessionCodes helper to check whether the given `resourceId` is an accession code or a dataset ID (deb7269a9f2a098b8b865c6bb3de6b9e9b55e189).

src/hooks/useResourceLoader.js
```diff
@@ -0,0 +1,80 @@
+import { useEffect, useState } from 'react'
+import { useDatasetManager } from 'hooks/useDatasetManager'
+import { useRefinebio } from 'hooks/useRefinebio'
+
+// resourceId: a processing dataset ID || a processing experiment accession code for the one-off experiment
+export const useResourceLoader = (resourceId, oneOffExperiment = false) => {
+  const { getDataset } = useDatasetManager()
+  const { processingResources, setProcessingResources } = useRefinebio()
+  const [hasError, setHasError] = useState(false)
```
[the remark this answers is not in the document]
nozomione: As mentioned in the [comment](https://github.com/AlexsLemonade/refinebio-web/pull/294#discussion_r1426015939), this implementation (which makes use of the `hasError` state) also matches the pattern of other hooks, so let's create a separate issue to refactor this hook pattern to make the error state as part of a context 👍

src/hooks/useDatasetManager.js
```diff
@@ -89,7 +89,9 @@ export const useDatasetManager = () => {
   }
 
   const startProcessingDataset = async (id, options) => {
-    const isCurrentDatasetId = id === datasetId
+    const isCurrentDatasetId = id && id === datasetId
+    const datasetIdToUse =
+      !id || !isCurrentDatasetId ? await createDataset() : id
```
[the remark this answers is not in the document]
nozomione:
```javascript
/* useDatasetManager.js */
// it takes parameters for a dataset ID and download options
  const startProcessingDataset = async (id, options) => { … }
```

This method is also used for one-time experiment downloads. 

Since an one-off has no associated dataset ID by default, `null` was passed as the first argument  `id`.  For this, the `!id ||` check was also added in this PR - commit at f8291c70385c840c72e479a0b4e52cd452a2f33d.

To revert this commit change, I’ve updated the parameter order and made `id`  an optional parameter with its default value (as below), removed the extra logical OR, and adjusted argument order in the files that use this method (6f99fd837d813739aa7ff33f59a048bb2f562ede).

```javascript
/* useDatasetManager.js */
  const startProcessingDataset = async (options, id = null) => { … }
```

nozomione:
@davidsmejia this PR is ready for another look.

src/hooks/useResourceLoader.js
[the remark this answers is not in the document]
nozomione: *also thank you for this great idea!

davidsmejia (commented):
Left a comment about removing the error on polling from the dataset page header. We should also look at how the `usePollDatasetStatus` hook should handle distinguishing between one off downloads and the "my dataset" dataset.

It would make sense to me to move the use of the polling hook to the `useDatasetManager` to allow for polling on that state instead of creating an interface that takes either an experiment accession or a dataset id. That way we wouldn't have to do inference on the type of ID as we should only be polling once the dataset is created.

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  const isProcessingDataset = latestDatasetState?.is_processing
+
+  const refreshProcessingResource = async () => {
+    const { datasetId } = getProcessingResource(resourceId)
+    const response = await getDataset(datasetId)
+
+    setHasError(response?.ok === false)
+    setLatestDatasetState(response)
```
davidsmejia, line 65: We should only update this on successful fetches.

src/components/Dataset/DatasetPageHeader.js
```diff
@@ -46,7 +45,7 @@ export const DatasetPageHeader = ({ dataset }) => {
   const isProcessing = dataset?.is_processing
   const isProcessingError = dataset?.success === false // 'success' may be null
 
-  if (isProcessingError || error) {
+  if (isProcessingError || hasError) {
```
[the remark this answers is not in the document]
nozomione: Removed at 09fdc6e996d56ccf7c41cd47f6150221ca7eb8b0.

nozomione:
@davidsmejia I've applied the feedback and this PR is ready for another look!

nozomione:
> Let's remove the usage of "resource" in variable names and function names associated with this PR.

What term would you suggest in replace of `resource` ?

The `dataset` and `datasetId` are taken by `DatasetManager`, and we're also passing  an accession code to this hook, we want to use a name that represents both of those resources. Perhaps what do you think about `selectedDataset`?

> Break up the "getters" since we will know when using this hook if we want to check if an accession has a processing dataset or if we are checking by dataset id explicitly.

Currently the resource id check is done by using the regular expression:
https://github.com/AlexsLemonade/refinebio-web/blob/09fdc6e996d56ccf7c41cd47f6150221ca7eb8b0/src/hooks/usePollDatasetStatus.js#L49-L51

Could you elaborate on this one?

Thank you David!

davidsmejia:
> > Let's remove the usage of "resource" in variable names and function names associated with this PR.
> 
> What term would you suggest in place of `resource` ?
> 
> The `dataset` and `datasetId` are taken by `DatasetManager`, and we're also passing an accession code to this hook, we want to use a name that represents both of those resources. Perhaps what do you think about `selectedDataset`?
> 
> > Break up the "getters" since we will know when using this hook if we want to check if an accession has a processing dataset or if we are checking by dataset id explicitly.
> 
> Currently the resource id check is done by using the regular expression:
> 
> https://github.com/AlexsLemonade/refinebio-web/blob/09fdc6e996d56ccf7c41cd47f6150221ca7eb8b0/src/hooks/usePollDatasetStatus.js#L49-L51
> 
> Could you elaborate on this one?
> 
> Thank you David!

I think these questions are related so I am going to provide a single response to all of them first.

The `useDatasetManager` should be responsible for creating, querying, configuring, and describing a dataset. Essentially activity related to a specific dataset.

The `usePollDatasetStatus` is the collection of logic around polling a specific dataset. This currently, and correctly, uses `useDatasetManager` under the hood to perform the query.

The change I would like to see here is how we associate a specific accession code with a dataset.

Currently `usePollDatasetStatus` does the following in this PR:
- Initializes the hook with either an accession code or a dataset id.
- fetches the latest state on mount
- provides helper methods for managing a list of currently polled datasets, either by ID of accession code.
- stores an array of objects that contain datasetid and optionally accession code for association

I would like to see the ambiguous argument (dataset id / accession code) removed from the initialization of the hook and from the `add.

We would accomplish this by:
- removing the initialization argument from the `usePollDatasetStatus` hook
- remove the `useEffect` that checks if the ID can be found and refreshes if it is lines 15-19
- move `processingDatasets` (old name `processingResources`) `useDatasetManager`
- Add a local storage state in RefinebioContext / exposed in hook with value of `datasetAccessions` that will default to an empty dict. The keys will be accession codes and the values will be dataset ids.
- add new method `pollDatasetId`
- add new method `pollDatasetAccession`
  - these functions should just set the datasetid and that is what the useEffect/polling logic should check

nozomione:
The following updates were completed:
> - remove the `useEffect` that checks if the ID can be found and refreshes if it is lines 15-19
> - move `processingDatasets` (old name `processingResources`) `useDatasetManager`

And now that we've moved `processingDatasets` to `useDatasetManager`, I've also: 
- moved `addDatasetToProcessingDatasets`(previously `addProcessingDataset` from `usePollDatasetStatus`) to `useDatasetManager`
- call  `addDatasetToProcessingDatasets` from `startProcessingDataset` to handle the addition of a new processing dataset (no longer calling it from two separate locations)

(see commit at be12eb546c9d68058a5c506f981852100c7e0057)

For the following items, I have a few questions: 

**Q 1:** How do we determine which dataset ID to poll inside the hook if we remove the hook's parameter?
> - removing the initialization argument from the `usePollDatasetStatus` hook

<details>
   <summary><strong>Reasons for passing an ID as an argument to <code>usePollDatasetStatus</code> hook</strong></summary><br/>

- To quickly check the given ID and initiate a polling request on component mount 
- To keep the polling logic inside the hook as much as possible (including ID check, start/refresh/end polling)

```js 
usePollDatasetStatus(processingId) // either a dataset ID or an one-off experiment accession code
```

When the following UIs are viewed by the user (on component mount), this hook checks the given ID, starts polling requests, and display the processing UI while it's being processed:

- Search card and its download modal in the search results page (passing an accession code to the hook):
Screenshot 2024-06-03 at 5 14 09 PM

- Experiment page for a specific accession code (passing an accession code to the hook):
Screenshot 2024-06-03 at 5 14 20 PM

- Dataset page that renders a specific dataset ID (passing a dataset ID in the URL to the hook):
(e.g., `/dataset/695fd15-2f7c-4b5c-8b39-212f9ee7b21e`)
Screenshot 2024-06-03 at 5 15 06 PM

<hr/>

</details>

By removing the argument for the hook (ID to check) and adding the next two methods:
> - add new method pollDatasetId
> - add new method pollDatasetAccession
>   - these functions should just set the datasetid and that is what the useEffect/polling logic should check

**Q 2:**  Where would you like the dataset ID to be set and start the polling request? 

To put another way, if this hook no longer accepts an ID as an argument, then at first, the ID check must to be done outside of the hook (a component that uses this hook by calling these new methods, `useEffect` etc). And possibly adding a new state to the hook to store that obtained dataset ID so that we can use it inside the hook to start polling. 

I suggest preserving the hook's argument to maintain the quick ID check and the polling logic within the hook as much as possible.

I'd like to hear your input on this. Please let me know, thank you David!

davidsmejia:
Ok, I think your questions stem from the same point so let me try clarifying this first and if still unclear I will try to tackle the questions separately.

Add a local storage state in RefinebioContext / exposed in hook with value of datasetAccessions that will default to an empty object. The keys will be accession codes and the values will be dataset ids.

So you don't have to care that it is a valid accession code here. When you add the dataset with optional accession code, you populate a separate object that functions like a look up table. No need for the regex.


So update what you have in `useDatasetManager` starting at line 34
Here you only need to assign a flat value of processing datasets and an object to look up by accession code.
So later in the application you will always know one of two things when you need to check if a dataset is processing:
- The dataset id, this is obvious
- The accession code which can be looked up to get the dataset id
 
`useDatasetManager.js`
```javascript
 const addToProcessingDatasets = (id, accessionCode) => {
    if (accessionCode) {
      setDatasetAccessions({...datasetAccessions, [accessionCode]: id})
    }
    setProcessingDatasets((prev) => {
      if (prev.includes(id)) return prev
      return [...prev, id]
    })
  }
```

Then you want to poll for a specific dataset. This should be done by calling one of the two functions below:
`usePollDatasetStatus.js`
```javascript

// this is the state variable that will trigger a rerender when updated
const [polledDatasetId, setPolledDatasetId] = useState(null)
const [datasetState, setDatasetState] = useState(null)
const timerRef = useRef(null)


// check if the hook should poll a dataset
  useEffect(() => {
    if (polledDatasetId) {
      timerRef.current = setInterval(() => {
        refreshProcessingDataset()
      }, 1000 * 60)
    }
    return () => {
      clearInterval(timerRef.current)
    }
  }, [polledDatasetId])

// by dataset ID 
pollDatasetId(datasetID) {
  // the dataset doesnt need to be polled anymore because it has been removed
  if (processingDatasets.includes(datasetID)) {
    setPolledDataset(datasetID)
  }
}

// tell the hook to watch a specific accession
/// call the above method if there is a match
pollDatasetAccession(accession) {
  if (datasetAccessions.hasOwnProperty(accession)) {
    pollDatasetId(datasetAccessions[accession])
  }
}

```

Let me know if you can see how this would work. Each hook would only be responsible for tracking the state of a single dataset. We should have a separate cleanup function ie `removeProcessingDataset` that only takes dataset id and removes it from the list of processing datasets and the datasetAccessions object.

nozomione:
The following updates were completed:
> - removing the initialization argument from the usePollDatasetStatus hook
> 

Now the hook no longer takes an ID argument and instead uses a new state, `polledDatasetId`, to set the processing dataset ID.

(see commit at 0d8ac185cfeb6566a552ba97471d3b7dd23c8ac9)

> - Add a local storage state in RefinebioContext / exposed in hook with value of datasetAccessions that will default to an > empty dict. The keys will be accession codes and the values will be dataset ids.

A lookup table `datasetAccessions` local storage state has been added to `RefinebioContext` to store the accession code of processing one-off experiment and imported into `useDatasetManager` used by `usePollDatasetState`. 

(see commit at b204a5da3c5de25e55ddaa8f0cccc44ca1aa95c3)

> - add new method pollDatasetId
> - add new method pollDatasetAccession
>    - these functions should just set the datasetid and that is what the useEffect/polling logic should check

The new methods have been added and are called from `useEffect` of components that uses the `usePollDatasetState` hook to set `polledDatasetId` for polling.  

(see previous commit and at b204a5da3c5de25e55ddaa8f0cccc44ca1aa95c3)

> a separate cleanup function ie removeProcessingDataset that only takes dataset id and removes it from the list of processing datasets and the datasetAccessions object.

The clean up method has been added which removes `polledDatasetId` from:
- the `processingDatasets` array
- the `datasetAccessions` object

Please note that I've also included a line that resets the `polledDatasetId` to `null` to stop the running polling timer (initially added it to the cleanup block of `refreshProcessingDataset`, but moved it to this method).

This PR is ready for your review, thank you David!

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+    datasetAccessions,
+    setDatasetAccessions,
+    processingDatasets,
+    setProcessingDatasets,
+    getDataset
+  } = useDatasetManager()
+  const [polledDatasetId, setPolledDatasetId] = useState(null)
+  const [polledDatasetState, setPolledDatasetState] = useState(false)
+  const timerRef = useRef(null)
+  const isProcessingDataset = polledDatasetState?.is_processing
```
davidsmejia, line 15:
```suggestion
  const [isProcessing, setIsProcessing] = useState(false)
```
If this is the only output we care about let's keep this a state and have `refreshProcessingDataset` handle setting this to true or false accordingly.

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+export const usePollDatasetStatus = () => {
+  const {
+    datasetAccessions,
+    setDatasetAccessions,
+    processingDatasets,
+    setProcessingDatasets,
+    getDataset
+  } = useDatasetManager()
+  const [polledDatasetId, setPolledDatasetId] = useState(null)
+  const [polledDatasetState, setPolledDatasetState] = useState(false)
```
davidsmejia, line 13:
```suggestion
```

We can remove this since we only really care about the status.

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+  const refreshProcessingDataset = async () => {
+    const response = await getDataset(polledDatasetId)
+    // TEMP: until the fetchAsync is refactored
+    if (response?.ok !== false) {
+      setPolledDatasetState(response)
+    }
+
+    if (!response.is_processing) {
+      removeProcessingDataset()
+    }
```
davidsmejia, line 69:
```suggestion
    const newIsProcessing = response.is_processing !== false
    setIsProcessing(newIsProcessing)
    if (newIsProcessing) clearInterval(timerRef.current)
```

This should be enough here. All we want to know if is the dataset is still processing. The `useDatasetManager` will handle putting and removing the dataset from the processing list. We just need to update the current state after fetching and then clear the timer if we no longer need it. Arguably, this could be put in another useEffect but this is fine as is in my opinion.

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
```diff
@@ -15,16 +15,21 @@ export const SearchCardAction = ({
   organismNames,
   technology
 }) => {
-  const { getProcessingExperiment } = useOneOffExperiment(accessionCode)
   const pageRendered = usePageRendered()
+  const { datasetAccessions, pollDatasetAccession } = usePollDatasetStatus()
```
davidsmejia, line 19:
```suggestion
  const { isProcessing, pollDatasetAccession } = usePollDatasetStatus()
```

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+  // if one-off, sets the matched accession code's dataset ID to polledDatasetId
+  const pollDatasetAccession = (accessionCode) => {
+    if (Object.hasOwn(datasetAccessions, accessionCode)) {
+      pollDatasetId(datasetAccessions[accessionCode])
+    }
+  }
+
+  // sets the mached dataset ID to polledDatasetId
+  const pollDatasetId = (datasetId) => {
+    if (processingDatasets.includes(datasetId)) setPolledDatasetId(datasetId)
```
davidsmejia, line 38: If you also update `isProcessing` to true when the dataset id is in `processingDatasets`, then the exported `isProcessing` will trigger a render and display correctly for the time before the first interval is expired.

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+  } = useDatasetManager()
+  const [polledDatasetId, setPolledDatasetId] = useState(null)
+  const [polledDatasetState, setPolledDatasetState] = useState(false)
+  const timerRef = useRef(null)
+  const isProcessingDataset = polledDatasetState?.is_processing
+
+  // polls the latest state of the processing dataset per minute
+  // (the processing usually takes a few minutes)
+  useEffect(() => {
+    if (polledDatasetId) {
```
davidsmejia, line 20: should eagerly clear the interval if `polledDatasetId` is true. This would prevent creating a new interval and losing the reference to the old one.

davidsmejia (commented):
Some comments around organization and improvement to the functionality of the hook.

The overall usage of the `usePollDatasetStatus` hook should be:

```javascript

const { isProcessing, pollDatasetAccession } = usePollDatasetStatus()

pollDatasetAccession(accession)

// logic using current state of `isProcessing`
```

src/hooks/usePollDatasetStatus.js
nozomione: Great, I used the separate `useEffect` as you suggested, see at bf3e091b238d595026c3e96e7a3ce86aa5b81f4d!

src/hooks/usePollDatasetStatus.js
```diff
@@ -0,0 +1,81 @@
...
+        const temp = { ...prev }
+        delete temp[keyToFRemove]
+
+        return temp
+      })
+    }
+
+    setProcessingDatasets((prev) => prev.filter((id) => id !== datasetId))
+    setPolledDatasetId(null) // removes polledDatasetId to stop the running timer
+  }
```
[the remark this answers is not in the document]
nozomione: I've also added a new method  `getProcessingDatasetByAccession` which returns a processing dataset ID in the `datasetAccessions` look up table by an accession code which is used in `ProcessingDatasetPillModal` (see at a5bf82650a3adc2e4925149acbcb0b3b4820ee21) as chatted in 1:1.

src/hooks/usePollDatasetStatus.js
nozomione: I've removed this local variable and added a new state `isProcessingDataset` (boolean), see at a5bf82650a3adc2e4925149acbcb0b3b4820ee21.

src/hooks/usePollDatasetStatus.js
nozomione: This `polledDatasetState` state is currently used to re-render `DatasetPageHeader`, please see the code block at aee530bbca025798b987282da8e92a45f2369c36.

src/hooks/usePollDatasetStatus.js
nozomione: The new state  was added at a5bf82650a3adc2e4925149acbcb0b3b4820ee21.

nozomione: See update at a5bf82650a3adc2e4925149acbcb0b3b4820ee21.

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
```diff
@@ -1,7 +1,7 @@
-import { memo } from 'react'
+import { memo, useEffect } from 'react'
```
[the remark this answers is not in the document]
nozomione: Without `useEffect` and calling `pollDatasetAccession` will results in run-time error (`Too many re-renders`). 

Please see Vercel preview [here](https://refinebio-gr8s0h2ob-ccdl.vercel.app/search) (at 5804932e336bef1d6875eca092d248b44d26a1cc).

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
```diff
@@ -15,16 +15,21 @@ export const SearchCardAction = ({
...
   const hasRnaSeq =
     typeof technology === 'string'
       ? technology === rnaSeq
       : technology.find((x) => x === rnaSeq)
+  const processingExperiment = datasetAccessions[accessionCode]
+
+  useEffect(() => {
+    // sets a processing datasets for polling
+    pollDatasetAccession(accessionCode)
+  }, [accessionCode, datasetAccessions])
```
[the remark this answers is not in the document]
nozomione: Please see https://github.com/AlexsLemonade/refinebio-web/pull/294#discussion_r1630376252.

nozomione:
I've updated the PR based on your feedback and ready for your review, thank you David!

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
nozomione: Now using the new state `isProcessingDataset`, see at a5bf82650a3adc2e4925149acbcb0b3b4820ee21.

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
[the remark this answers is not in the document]
davidsmejia: You can fix that by just adding a guard on `pollDatasetId`. Just return early if the passed in `datasetId` is already `polledDatasetId` since you can presume you already have a timer for that in place.

davidsmejia (commented):
Responded to your question and added some suggestions for clarity.

src/hooks/useDatasetManager.js
```diff
@@ -100,10 +151,11 @@ export const useDatasetManager = () => {
...
     }
-    const response = await updateDataset(
-      isCurrentDatasetId ? id : await createDataset(),
-      params
-    )
+
+    const datasetIdToUse = isCurrentDatasetId ? id : await createDataset()
+    const response = await updateDataset(datasetIdToUse, params)
+    // adds this dataset ID to processingDatasets[] for polling
+    addToProcessingDatasets(datasetIdToUse, accessionCode)
```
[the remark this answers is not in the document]
nozomione: See commit at e65aa12dff361a3983717deb1d28c03a937fe8cc.

src/hooks/usePollDatasetStatus.js
nozomione: See update at 20607db98fa01e5661ccbbc8704d7c1f132774c5.

src/hooks/useDatasetManager.js
```diff
@@ -83,13 +124,23 @@ export const useDatasetManager = () => {
       setDataset(formattedResponse)
     }
 
+    // removes this dataset ID from processingDatasets[] if it exists
+    if (!response.is_processing) {
```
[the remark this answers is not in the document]
nozomione: See update at 3bb2858f3df47a6c092cec26347139f5407984ee.

src/components/shared/SearchCard/SearchCardAction/SearchCardAction.js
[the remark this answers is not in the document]
nozomione: See update at 89f9daea7d91711677556287cf635924d38a667d.

src/hooks/useDatasetManager.js
```diff
@@ -26,7 +30,44 @@ export const useDatasetManager = () => {
...
+      (k) => datasetAccessions[k] === id
+    )
+    if (keyToFRemove) {
+      setDatasetAccessions((prev) => {
+        const temp = { ...prev }
+        delete temp[keyToFRemove]
+
+        return temp
+      })
+    }
```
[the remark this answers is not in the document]
nozomione: See updates at 4b4ceaf8a91c7f48b5c73ab13fcbb44c87aaa3cf and a41eb4ed1eb226a88f47006fa20f02c37f42df64.

src/hooks/useDatasetManager.js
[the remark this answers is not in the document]
nozomione: See update at e65aa12dff361a3983717deb1d28c03a937fe8cc.

src/pages/dataset/[dataset_id].js
Comment (davidsmejia) [inline src/pages/dataset/[dataset_id].js:1]: Before we merge this, we need to clean this file up. I should be able to look at this and tell which dataset we are presenting here. As far as I can tell we should *only* care about the dataset from the query parameter. Please explain if I am missing something.

Comment (nozomione) [reply src/pages/dataset/[dataset_id].js:1]: This page handles two types of datasets:

**Type 1:** `selectedDataset` - the dataset currently selected on the page

This state is used to store the dataset currently displayed on the page, which is the dataset ID obtained from the URL query parameter stored in `idFromQuery`.

<details>
  <summary>A dataset ID from URL</summary><br/>

A dataset ID `0a226474-fc40-4fe1-bc32-836ccb788b11` in URL:

idFromQuery

<hr/>
</details>

**Type 2:** `regeneratedDataset` - the dataset to be regenerated (see #287 for reference)

This state is used to store a modified version of the already processed dataset. Due to the immutability of the processed datasets, a new dataset ID will be generated if a user edits the download options before regenerating the processed dataset displayed on the page (`selectedDataset`). 

And we use this newly created dataset to re-render the modified dataset details of the processed dataset in the Download Files Summary section for that user.

<details>
  <summary><strong>Example</strong></summary><br/>

To replicate: 

**Step 1:** Visit https://refinebio-c1amq1arv-ccdl.vercel.app/dataset/0a94212a-04bf-4733-a3f6-0d9caed5a0bb

**Step 2:** Change the processed dataset's download options to see the UI update

Before updating the download options:

BeforeUpdatingDownloadOptions

After updating the download options, a new dataset will be created behind the scene and stored in `regeneratedDataset`. We'll then swap the dataset currently displayed (`selectedDataset`) only in the Download Files Summary section to this new one to reflect the new changes in the UI accordingly:

Screenshot 2024-06-13 at 5 12 44 PM

<hr/>
</details>

**Other variables:**

 `isSameId` is used for the Move to Dataset button state. If the dataset ID currently displayed on the page is the same as My dataset ID (`datasetId`), then this button is disabled, otherwise is enabled. 

<details>
  <summary><strong>Example</strong></summary><br/>

To replicate: 

**Step 1:** Add an experiment to My Dataset

**Step 2:** Click the Share Dataset button to get a URL

**Step 3:** Open the URL in the browser

e.g.) If the dataset ID currently displayed is same as My dataset, we disable the Move to Dataset button:

isSameId

<hr/>
</details>

`isUnprocessedDataset` is used to display the Download Dataset button. If the dataset currently displayed on the page is unprocessed (i.e. unprocessed shared dataset), we display this button in the UI, otherwise hide it.

<details>
  <summary><strong>Example</strong></summary><br/>

For unprocessed, the Download Dataset button is visible: 
https://refinebio-c1amq1arv-ccdl.vercel.app/dataset/7b501023-98f6-4f0f-b410-87fe4a6e5d19?ref=share

ForUnprocessedDataset

For processed dataset, it's hidden: 
https://refinebio-c1amq1arv-ccdl.vercel.app/dataset/0a226474-fc40-4fe1-bc32-836ccb788b11

ForProcessedDataset

<hr/>
</details>

 `isProcessed` is used to display the download options in the Download Files Summary section to regenerate the processed dataset.

<details>
   <summary><strong>Example</strong></summary><br/>

For processed dataset, the download options is visible: 
https://refinebio-c1amq1arv-ccdl.vercel.app/dataset/0a226474-fc40-4fe1-bc32-836ccb788b11

downloadOptionVisible

For unprocessed, it's hidden:
https://refinebio-c1amq1arv-ccdl.vercel.app/dataset/7b501023-98f6-4f0f-b410-87fe4a6e5d19?ref=share

downloadOptionHidden

<hr/>
</details>
  
Please let me know if you have any feedback or questions, thank you David!

davidsmejia (approved):
Ok, that page should be reworked so that it isnt juggling variables like that. But we should handle that in a different issue and PR.

For the issue this PR is good for me

Issue #270, opened 2023-11-08
Implement error handling for API resource requesets for datasets on the dataset and download pages


nozomione:
### Context
We currently configured error handling in `_app` using Sentry-integrated React.js [`ErrorBoundry`](https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/) with our custom error component which handles server-side errors `500` and `504` and a runtime error `404` at the page level. We'll also need to implement error handling for API resource requests and display an appropriate error view in a page based on the status code returned by the server.

This issue is to implement error handling for the dataset endpoint (for the dataset and download (My Dataset) pages).

### Problem or idea
We should implement error handling for dataset resource requests and render an appropriate error view in the dataset and download (My Dataset) pages based on the error status code returned by the server.

### Solution or next step
(*due to frequent PR merges, the most recent PR's feature branch will be used as the base branch to reduce merge conflicts)

- Create or reuse a component for the dataset endpoint error handling 
- Create any helpers as necessary
- Update the existing error handling logic if required

PR #297, opened 2023-12-04
270 - Implement error handling for API resource requests for datasets on the dataset and download pages

nozomione:
## Issue Number
Closing #270

Implement error handling for the dataset and download pages.

Please see the latest UI [here](https://refinebio-web-git-nozomione-270-implement-error-han-937bb5-ccdl.vercel.app).

## Types of changes

## Checklist


src/hooks/useDatasetManager.js
```diff
@@ -74,6 +74,14 @@ export const useDatasetManager = () => {
           }
         : {}
     const response = await api.dataset.get(id || datasetId, headers)
+
+    if (response?.ok === false) {
+      setError({
+        hasError: true,
+        statusCode: response.statusCode
+      })
+    }
```
davidsmejia, line 83: I'm not sure we want to handle errors this way. We are conditionally setting a nested attribute to false when an error occurs. I don't see this an an improvement over setting a single atomic state variable as either a status code or null and interpreting error from there.

I think something like this would be more appropriate
```suggestion
    const { ok, statusCode } = response
    if (error !== statusCode ) {
      if (ok ===  false) setError(response.statusCode)
      if (ok && statusCode) setError(null)
    }
```

nozomione: See the commit b39738f4d083c7b99be149a307824ceda204f2b6. 

I also left `TODO` comment for `ok && error` check (see this related PR #337 for reference), thank you!

nozomione:
I've updated this PR based on your input and is now ready for your review, thank you David!

src/hooks/useDatasetManager.js
nozomione: **NOTE:** PR #337 was created based on the talk regarding the `ok` property https://github.com/AlexsLemonade/refinebio-web/pull/294#discussion_r1466558333 (and first mentioned here https://github.com/AlexsLemonade/refinebio-web/pull/294#issuecomment-1911167634). This is because it requires a small refactoring of other part of the codebase (this was back in January and was old thread so wanted to add more context), thank you David!

src/hooks/useDatasetManager.js
```diff
@@ -27,7 +27,7 @@ export const useDatasetManager = () => {
     token
   } = useContext(DatasetManagerContext)
   const { createToken, resetToken, validateToken } = useToken()
-  const [error, setError] = useState(false)
+  const [error, setError] = useState(null) // stores an error status code, otherwise null
```
davidsmejia, line 30: I don't think the comment in necessary here.

src/hooks/useDatasetManager.js
```diff
@@ -115,6 +115,18 @@ export const useDatasetManager = () => {
...
+    const { ok, statusCode } = response
+    // sets the error if any, otherwise resets it
+    // TODO: add a check (ok && error) to default the error state
+    // related created PR https://github.com/AlexsLemonade/refinebio-web/pull/337
+    if (error !== statusCode) {
+      if (ok === false) {
+        setError(statusCode)
+      } else {
+        setError(null)
+      }
```
davidsmejia, line 127:
```suggestion
      setError(ok ? null : statusCode)
```

davidsmejia (approved):

src/hooks/useDatasetManager.js
nozomione: Completed at 009323f324e84e42fc5cc47670ea1b3d32e92db3.

src/hooks/useDatasetManager.js
nozomione: Completed at 009323f324e84e42fc5cc47670ea1b3d32e92db3.

Issue #209, opened 2023-10-10
Initial GA4 Analytics Configuration


davidsmejia:
### Context

Epic: https://github.com/AlexsLemonade/refinebio-web/issues/302

We need to do some basic tracking to understand how users interact with refinebio-web.

### Problem or idea

We want to initially look at default report generated by just dropping in GA4 and then doing internal site-wide testing.
We should limit the accompanying PR to only contain the addition of [react-ga4](https://www.npmjs.com/package/react-ga4) and the requisite configuration.

A future Issue will be made to track if we should create a hook and individual tracking elements.

### Solution or next step
Edited by @nozomione 
- [x] Provide GA4 measurement IDs @davidsmejia   
- [x] Set up `production` and `staging` environmental variables in the next config and Vercel @nozomione 
   - `GA4_MEASUREMENT_ID`
   - `STAGE_GA4_MEASUREMENT_ID`
- [x] GA4 Integration @nozomione 
     - (approach 1) Add GA4 script to the project 
     -  ~~(approach 2) Install a project dependency [`react-ga4`](https://github.com/codler/react-ga4) and pass GA4 measurement ID~~

 **NOTE** The approach 1 was taken and the [gtag](https://developers.google.com/analytics/devguides/collection/ga4/events?client_type=gtag) function can be used to set up our custom events, so the `react-g4` package was not added.

davidsmejia:
The code for analytics can  now be found in 1password.

PR #296, opened 2023-11-21
295 - Clean up docker-compose.env

nozomione:
## Issue Number

Closing: #295

## Purpose/Implementation Notes
Added the following env variables to `docker-componse.env` and adjusted the codebase accordingly:

|Name|Value|Note|
|:---|:---|:---|
|`STAGE_SENTRY_DSN`| `''`|Always an empty string|
|`STAGE_SENTRY_ENV`| `local-refinebio-client`|constant unless changed|
|`STAGE_API_HOST`| `STAGE_API_HOST`|`staging` API host name|
|`STAGE_API_VERSION`|STAGE_API_VERSION|current `staging` API version|

## Types of changes

What types of changes does your code introduce?

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Functional tests

Ran all the scripts locally and successfully started/removed the containers for `client` and `storybook`.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #305, opened 2023-12-21
209 - Initial GA4 Analytics Configuration 

nozomione:
## Issue Number
Closing #209

Added Gtag (for GA4) scripts to the project  for `staging` and `production`.

(**Note:** Instead of using [`react-g4`](https://github.com/codler/react-ga4) package, the [gtag](https://developers.google.com/analytics/devguides/collection/ga4/events?client_type=gtag) function will be used to set up our custom events.)

## Types of changes

## Checklist


src/pages/_app.js
```diff
@@ -15,6 +16,9 @@ import { PageTitle } from 'components/shared/PageTitle'
 
 getPageLoader()
 const Fallback = () => <ErrorPage />
+const GA4MeasurementID =
+  process.env.GA4_MEASUREMENT_ID ||
+  process.env.NEXT_PUBLIC_DEV_GA4_MEASUREMENT_ID
```
davidsmejia, line 21: We don't need to do this. We should be setting the `STAGE_GA4_MEASUREMENT_ID` in `docker-compose.env`

nozomione: Based on the merged PR #296,  the `docker-compose.env` file is now excluded from commits. 

For local development, please note that the measurement ID used is the dev GA property `CCDL > Refine.bio Web - Dev` in our analytics account. Thank you David!

davidsmejia (approved):

PR #301, opened 2023-12-11
223 - Integrate Github API to the request data form for filing issues in refinebio repo

nozomione:
## Issue Number
Closing #223

## Purpose/Implementation Notes
- Temporarily using a [test repo](https://github.com/nozomione/test-dataset-request/issues) for local development, but the correct values (endpoint and token)  for `staging` and `production` have been added to Vercel.

---
Integrated the Github API to the request data form for creating an issue in `refinebio` repo:
  - Request Search Data Form via Search (terms with no search results)
  - Request Experiment Data Form via the "Request Experiment" button (experiments with non-downloadable samples) 

<ins>Via Request Search Data Form</ins>
When a user makes a data request, the application redirects to the homepage and displays a notification banner based on the request status (Error = GitHub and/or HubSpot failed, Success = both successful).

Success banner:

Screenshot 2023-12-11 at 12 59 23 PM

Error banner (also the failed data request will be posted in Slack CCDL channel):

Screenshot 2023-12-11 at 1 09 35 PM

<ins>Via Request Experiment Data Form</ins>
If the data request was successful, the application marks the experiment as requested in the search card, otherwise the failed data request will be posted in Slack CCDL channel:

Screenshot 2023-12-11 at 1 13 59 PM

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

Issue #280, opened 2023-11-13
QA: Anchor View samples button to samples section on experiment page


dvenprasad:
### Context
When I click on the `View Samples` button, it takes me to the top of the experiment page.

### Problem or idea
It should link to the samples section on the experiment page instead

### Solution or next step

Tagging @davidsmejia for next steps

nozomione:
Hello @dvenprasad, This behavior is the same as the current refine.bio. The data in the samples table needs to be fetched first when the experiment page loads. For this, it displays a loading spinner animation when a user enters the page.

Do you want to add an auto-scroll-down to the samples table section **after** the data has finished loading? Please let me know!

nozomione:
Yes, it typically doesn't take more than a few seconds! The page renders a loading spinner while fetching the submitter supplied information and the samples table section. The data in the samples table is loaded using a separate hook. The page will render the table UI once the data is loaded (with populated HTML element's properties with that fetched data) and auto-scroll it to the top of the page.

PR #306, opened 2024-01-04
280 - QA: Anchor View samples button to samples section on experiment page

nozomione:
## Issue Number
Closing #280

(This branch was created based on PR https://github.com/AlexsLemonade/refinebio-web/pull/301.)

Added the auto-scroll animation to the samples table section, when a user clicks the "View Samples" button to access the experiment page via the search page. The UI scrolls the samples table section to the top of the page as the data is loaded.

Screenshot 2024-01-03 at 7 33 35 PM

Please see the latest UI [here](https://refinebio-k72hn2i5o-ccdl.vercel.app/search).

## Types of changes

## Checklist


dvenprasad (approved):
Looks good! 
🦓 🦓 🦓

src/contexts/RefinebioContext.js
```diff
@@ -22,6 +22,7 @@ export const RefinebioContextProvider = ({ children }) => {
     'requested-experiments',
     []
   )
+  const [samplesTableData, setSamplesTableData] = useState({})
```
davidsmejia, line 25: I'm having trouble understanding why we should put this data here. This is our global application context and should be used for data that needs to be persisted and accessed in multiple locations.

davidsmejia (commented):
Left a question about approach. Do you think that `scrollView` would be better served written as a hook?

src/contexts/RefinebioContext.js
nozomione: The `samplesTableData` was previously shared with `src/pages/experiments/[accession_code]/[experiment]/index` to calculate the correct scroll position of the samples table scroll animation (ensuring the table data was fully loaded). But it has been moved from the global context and is defined in `SamplesTableContext` (at cee5282ac17f02756443223a717fb6cb5a07cdd5).

Instead, I am now using the [ResizeObserver API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to handle the calculation for the samples table position on layout changes (at f5edd0278fd6caed0e1b947db07e8f9b6e8b3c26).

I've also made a small revision to `helpers/scrollTop` (previously was named `scrollToTop`) and adjusted the codebase (at ec71236b7c3577d370dff6239eb7e0f9f4aa2f2f). Thank you, David!

davidsmejia (commented):
Can we first clean up the renaming of context hook state getters/setters.

nozomione:
I've updated this PR and it's ready for another look. Thank you David!

nozomione:
As discussed in 1:1, I've replaced the `querySelector` method with the Context API.

It now handles access to the `<header />`  element using `LayoutRefsContext` and its hook `useLayoutRefs`  (dde79d12fa284fee0a3a7d5d4a301f6c8b80a38c). Additionally, I've cleaned up the `sampleTablesData` states.

Also I am going to create a separate issue for further refactoring of `useSamplesTableManager` for the next sprint, as we discussed 👍 

This PR is ready for your review, thank you David!

Issue #303, opened 2023-12-13
Implement GA4 custom events script


nozomione:
### Context
Epic: https://github.com/AlexsLemonade/refinebio-web/issues/302

We need to implement a script for GA4 custom events to collect and analyze our site data.

Docs for required data collection and custom events: 
Slab: [Analytics Strategy](https://data-lab-knowledge.slab.com/posts/refine-bio-analytics-strategy-cppx30qp)
Google sheet [Custom events](https://docs.google.com/spreadsheets/d/1qKeq8_pEezvq-OAp6ohCjJhrr9EQYLLyGYFAQe9XvS4/edit#gid=1911374202) 

### Problem or idea
We should implement the script for GA4 custom events to track our application data listed in above.

(**Note:** [gtag](https://developers.google.com/analytics/devguides/collection/ga4/events?client_type=gtag) API was used (is available to use when GA script loads) to implement custom events instead of [`react-g4`](https://github.com/codler/react-ga4) package to reduce the dependencies.)

### Solution or next step
Create the custom events script that contains all methods required for data collection ([sheet](https://docs.google.com/spreadsheets/d/1qKeq8_pEezvq-OAp6ohCjJhrr9EQYLLyGYFAQe9XvS4/edit#gid=1911374202)).

PR #321, opened 2024-01-22
303 - Implement GA4 custom events script

nozomione:
## Issue Number
Closing #303

Created the GA4 custom events script  with gtag API that contains all methods required for data collection ([sheet](https://docs.google.com/spreadsheets/d/1qKeq8_pEezvq-OAp6ohCjJhrr9EQYLLyGYFAQe9XvS4/edit#gid=1911374202)).

## Types of changes

## Checklist


src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+  event('my_dataset_download_options', {
+    my_dataset_download_options: `Aggregate: ${
+      dataset.aggregate_by
+    }, Transformation: ${transformation[dataset.scale_by]}, QN: ${
+      dataset.quantile_normalize ? 'Skipped' : 'Not skipped'
+    }
+  `
+  })
+
+const datasetDownload = (token) => event('dataset_downalod', { token })
```
davidsmejia, line 44:
```suggestion
const datasetDownload = (token) => event('dataset_download', { token })
```

Which token does this refer to?

src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+    one_off_experiment_download: accessionCode
+  })
+
+const regeneratedDataset = (state, defaultOptions, newOptions) => {
+  // due to GA char limit, keys names are abbreviated
+  const format = (option) => {
+    return `A: ${option.aggregate_by}, T: ${
+      transformation[option.scale_by]
+    }, QN: ${option.quantile_normalize ? 'Not skipped' : 'Skipped'}`
+  }
```
davidsmejia, line 57: We should use the same logic as for `myDatasetDownloadOptions`. Use the shorter of the two if you are limited by characters.

src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+
+  event('regenerated_dataset', {
+    regenerated_state: state ? 'Expired' : 'Valid',
+    regenerated_download_options: `${format(defaultOptions)} ${
+      newOptions ? `(new) ${format(newOptions)}` : ''
+    }`
+  })
+}
+
+const sharedDataset = (status) =>
```
davidsmejia, line 67: Is status going to be a boolean?

src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+      parameter = 'social_media'
+      break
+    default:
+      break
+  }
+
+  event('click', {
+    [parameter]: url
+  })
+}
```
davidsmejia, line 117:
```suggestion
const outboundClick = (url) => {
  const [linkName] = Object.entries(links).find(([key, value]) => value === url)
  const parameter = `click_to_${linkName}`

  event('click', {
    [parameter]: url
  })
}
```

Can we instead do something like this?

src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+  // due to GA char limit, keys names are abbreviated
+  for (const x of keys) {
+    if (filters[x]) {
+      formattedFilters = `${formattedFilters}[${
+        x === keys[0] ? 'O' : x[0].toUpperCase()
+      }]:${filters[x]} `
+    }
+  }
+
+  return formattedFilters
```
davidsmejia, line 150:
```suggestion
  // due to GA char limit, keys names are abbreviated
  return keys
   .filter((key) => filters[key])
   .map((key) => `[${key[0].toUpperCase()}]:${filters[key]}`)   
   .join(" ")
```

I'm a bit perplexed by this condition, I see from the comment that we are trying to truncate it. But the issue I see is if the first `key` is not in the filter, then we would never return `O` and always return `x[0].toUpperCase()` ?
```javascript
x === keys[0] ? 'O' : x[0].toUpperCase()
```

src/api/analytics/gtag.js
```diff
@@ -0,0 +1,192 @@
...
+  const [, o, p, t] = formatFacetNames(Object.keys(facets))
+  const filters = {}
+  for (const x of [o, p, t]) {
+    if (query[x]) {
+      filters[x] = sort({ [x]: query[x] })
+    }
+  }
+
+  return { keys: [o, p, t], filters }
+}
```
davidsmejia, line 173: I'm having some difficulty reading this. I think we should start by using some semantic variable names here.

I don't understand why we are dropping the first facet.

davidsmejia (commented):
I think this needs some work before another review. Could you review this yourself and see if it would be possible to make this a bit more clear. I have made some suggestions where I thought I understood what was trying to be accomplished.

src/api/analytics/gtag.js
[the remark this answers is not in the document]
nozomione: See at e8a9e5a5811997ce140b58f162027ceb43aa536e.

src/helpers/formatFacetNames.js
```diff
@@ -0,0 +1,18 @@
...
+
+  for (const name of facetNames) {
+    if (Object.keys(formattedFacetNames).includes(name)) {
+      formattedNames.push(formattedFacetNames[name])
+    } else {
+      formattedNames.push(name)
+    }
+  }
+
+  return formattedNames
```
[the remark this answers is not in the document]
nozomione: See at 98bea01de17eb9552a1ebc20d73546ab2e480a0d.

src/config/options.js
```diff
@@ -129,7 +129,12 @@ export const options = {
       label: 'Z-score',
       value: 'STANDARD'
     }
-  ]
+  ],
+  transformationHumanReadable: {
```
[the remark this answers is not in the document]
nozomione: I've reused `transformation` and removed `transformationHumanReadable`, see at ff27d84df24b985ea44f3896d6f28fb690b34e8e.

src/api/analytics/gtag.js
nozomione: I've renamed this param name to `isProcessed` for clearer indication of its data type at a5d1a15fff0ea82a489f79446cdc8501bbc23f5.

src/api/analytics/gtag.js
nozomione: Oh yes, this is because the property name for the downloadable organisms is returned as `downloadable_organism_names` from the BE API (https://api.refine.bio/v1/#operation/search_list) which is then shortened to `downloadable_organism` on the client.

In the UI, we display those downloadable organisms in the "Organism" section of the facets:
Screenshot 2024-08-15 at 6 52 05 PM

For this reason, if items from the "Organism" section are selected, it will be explicitly set to `O` (otherwise `D`) to match the UI.

e.g., )  The selected filters may be as follows:

```js
sortedFilters = {
      // the first item will be explicitly set to 'O' 
      keys: ['downloadable_organism', 'platform', 'technology'], 
      filters: {
        downloadable_organism: ['Homo sapiens'],
        platform: 'Illumina HiSeq 2500',
        technology: ['Microarray', 'RNA-seq']
      }
    }
 ```

In analytics, the above filter combination will be collected and displayed in the "Filter Combination" report as follows:

`[O]:Homo sapiens [P]:Illumina HiSeq 2500 [T]:Microarray,RNA-seq`

Filter Combination Report in `CCDL > Refine.bio Web -Dev` (to see the demo data, please set the custom date to `May 1, 2023` to present)

Screenshot 2024-08-15 at 7 43 22 PM



This data structure was discussed previously and added to the google sheet [Custom Events List > Search > Most used filter combination](https://docs.google.com/spreadsheets/d/1qKeq8_pEezvq-OAp6ohCjJhrr9EQYLLyGYFAQe9XvS4/edit?gid=1996653253#gid=1996653253) some times ago (back in January..!). When you have a moment, please take another look and let me know your thoughts on this. Thank you, David!

File: src/api/analytics/gtag.js (hunk 6 above)
Comment (nozomione) [reply src/api/analytics/gtag.js]: The reason for removing the first facet, `has_publication`, from `facets` is that it's not included in the data collection for the most used filter combination.

```js
  const [h, o, p, t] = formatFacetNames(Object.keys(facets))  
// h = "has_publication" <- this is unused thus dropped
// o = "downloadable_organism"
// p = "platform"
// t = "technology"
 ```
Initially, I used the first letter of each facet item (with `o` assigned to `downloadable_organism`), but I've now renamed them to the following for better clarity and semantics:
- `o` to `organismFacet`
- `p` to `platformFacet`
- `t` to `technologyFacet`  

Additionally, `x` is now renamed to `item`. I’ve also added a comment explaining the reason for removing the first facet (at 506fe4c7c0577784a3afe532efd1ce11be048204). Let me know your thoughts on this, thank you, David!

File: src/api/analytics/gtag.js (hunk 4 above)
Comment (nozomione) [reply src/api/analytics/gtag.js]: Thank you for the great suggestion! 👍  I’ve incorporated your advice with some tweaks and revised `outboundClick` (at 57de1f46b6c47d03a952742fa8a666387967bbf2).

I've added `specialCases` tweak primarily to avoid having long event parameter names (e.g., `click_to_refinebio_github_differential_expression_microarray` that is `60` chars). User-defined event parameters need to be registered as custom definitions in the GA property (e.g., `CCDL > Refine.bio Web - Dev`) to make sure they can be tracked and available in GA reports. Since the event parameter name limit is currently `40` chars [[GA4] Event collection limits](https://support.google.com/analytics/answer/9267744?hl=en), this tweak helps us keep the parameter name within that limit.

Please let me know your thoughts on this!

Based on this change, I’ve also updated the parameter names in the google sheet [Custom Events List > Outbound Link](https://docs.google.com/spreadsheets/d/1qKeq8_pEezvq-OAp6ohCjJhrr9EQYLLyGYFAQe9XvS4/edit?gid=1911374202#gid=1911374202) accordingly, thank you David!

File: src/api/analytics/gtag.js [hunk 9]
```diff
@@ -0,0 +1,192 @@
+import { links, options } from 'config'
+import formatFacetNames from 'helpers/formatFacetNames'
+import formatFilterName from 'helpers/formatFilterName'
+
+const { transformationHumanReadable: transformation } = options
```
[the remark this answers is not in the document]
nozomione: I've added a supporting comment at ff27d84df24b985ea44f3896d6f28fb690b34e8e which includes the data structure. Please let me know your thoughts on this, thank you David!

src/api/analytics/gtag.js
nozomione: I've converted to this `format` method into a new helper called `getFormattedDownloadOptions` and used it in both `myDatasetDownloadOptions` and `regenerateDataset` (at 2bb98c8c735e28b65a0f113aa3984fce1af84e56).

Btw, do you prefer a shorter name? (e.g., `formatDownloadOptions` or `getDownloadOptions`)?

src/api/analytics/gtag.js
nozomione: I've also renamed other parameters for consistency at 2bb98c8c735e28b65a0f113aa3984fce1af84e56, thank you David!

nozomione:
I’ve applied your feedback and also added comments to each function to describe what they do. This PR is now ready for another look, thank you David!

davidsmejia (commented):
This is partially reviewed and discussed. Let's apply the suggestions so far and revisit for the remainder of the items.

nozomione:
I've applied your suggestions at 87b47700f000b7342fe15593565e61e77b1b3159, which includes the following changes:

- Added new helper modules:
  -  `helpers/getReadable` 
  - `analytics/helpers`
- Added a new configuration file `config/readables` for `helpers/getReadable`
- Updated the `analytics/gtag` script to:
  - Accept the entire data structure instead of specific tracking values and extract necessary values inside each method
  - Define a new constant `payload` and assign extracted tracking values to it before passing it to GA `event` method
  - Utilize new helper modules 

This PR is ready for another look, thank you David!

src/analytics/gtag.js
```diff
@@ -0,0 +1,202 @@
...
+
+// tracks the dataset's state (expired or valid) and changes in download options (initial and updated)
+const regeneratedDataset = (isExpired, defaultOptions, newOptions) => {
+  event('regenerated_dataset', {
+    regenerated_state: isExpired ? 'Expired' : 'Valid',
+    regenerated_download_options: `${getFormattedDatasetOptions(
+      defaultOptions
+    )} ${newOptions ? `(new) ${getFormattedDatasetOptions(newOptions)}` : ''}`
+  })
+}
```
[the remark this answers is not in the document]
nozomione: Thank you for this great suggestion 👍, David!

src/analytics/helpers.js
```diff
@@ -0,0 +1,94 @@
...
+    return `${getReadable(key, dataset[key])} -> ${getReadable(
+      key,
+      regeneratedDataset[key]
+    )}`
+  })
+  return changes.join('|')
+}
+
+export const getDatasetState = (dataset) =>
+  getReadable('expires_on', moment(dataset.expires_on).isBefore(Date.now()))
```
nozomione, line 33: I've added the `expires_on` key to the config 's`readableBooleans` for `getReadable` (replacing the previous `isExpired` parameter). This key returns either `'Expired'` or `'Valid'` as fixed values. Let me know what you think!

```
export const readableBooleans = {
  expires_on: ['Expired', 'Valid'],
  quantile_normalize: ['Not skipped', 'Skipped'],
  quant_sf_only: ['rna-seq', 'normalized']
}
```

src/analytics/helpers.js
```diff
@@ -0,0 +1,94 @@
...
+export const event = (eventName, value = {}, nonInteraction = false) => {
+  window.gtag('event', eventName, {
+    ...value,
+    non_interaction: nonInteraction
+  })
+}
+
+// formats a given string to snake_case for event parameters
+export const getSnakeCase = (str) =>
+  str.replace(/([a-z])([A-Z])/g, '$1_$2').toLowerCase()
```
nozomione, line 20: I've added this utility helper method to `analytics/helpers` to convert component names from camelCase to snake_case that is the accepted format for GA event parameters. Should we keep it here or move it to `helpers`?

src/analytics/helpers.js
```diff
@@ -0,0 +1,94 @@
...
+    if (query[item]) {
+      filters[item] = sort({ [item]: query[item] })
+    }
+  }
+
+  return { keys: [organismFacet, platformFacet, technologyFacet], filters }
+}
+
+export const getToggledFilterItem = (isChecked, item) =>
+  `${isChecked ? 'Add' : 'Remove'} - ${item}`
```
nozomione, line 94: I've added this helper method to get the toggled filter item, including the fixed prefix values (`'Add' - ` and `'Remove - '`). I am thinking about adding this to the config `readableBooleans` and using `getReadable` to return these prefixes. What do you think?

src/analytics/gtag.js
```diff
@@ -0,0 +1,149 @@
...
+  event('regenerated_dataset', payload)
+}
+// tracks user clicks on the share dataset button
+const trackSharedDataset = () => {
+  event('shared_dataset')
+}
+
+/* --- Links --- */
+// tracks click-through to the experiment page from search results
+// (via title and "View Samples" button)
```
davidsmejia, line 76:
```suggestion
// tracks click-through to the experiment page from search results
// used to create a report to differentiate which component users click
```

src/analytics/gtag.js
```diff
@@ -0,0 +1,149 @@
...
+// tracks the outbound link clicks
+const trackOutboundClick = (url) => {
+  const linkName = getParameterForLink(url)
+  // no event will be sent if no match
+  if (!linkName) return
+
+  const payload = {}
+  payload[`click_to_${linkName}`] = url
+  event('click', payload)
+}
```
davidsmejia, line 103: These can have a single interface that is `trackLink` that should be called always in the `onClick` for an Anchor.

Then for the specific ones we wont be tracking the url just whatever we want for that report.

src/analytics/gtag.js
```diff
@@ -0,0 +1,149 @@
...
+  const payload = {}
+  payload.my_dataset_download_options = getFormattedDatasetOptions(dataset)
+  event('my_dataset_download_options', payload)
+}
+// tracks the number of dataset downloads associated with each token
+const trackDatasetDownload = (token) => {
+  const payload = {}
+  payload.token = token
+  event('dataset_downalod', token)
+}
```
davidsmejia, line 52: We don't need the token for this. Really shouldn't be sending that out.

src/analytics/helpers.js
nozomione: Yes, this is great 👍!

src/analytics/gtag.js
nozomione: We now use the following dimensions with their values set as URLs:
- `click_internal_link`
- `click_external_link`

This refactoring simplifies and narrows down the tracking items (and if needed, we can utilize filters in GA4 reports for more specific domain tracking) - great solution 👍!

src/analytics/helpers.js
```diff
@@ -0,0 +1,94 @@
...
+  for (const x of keys) {
+    if (filters[x]) {
+      formattedFilters = `${formattedFilters}[${
+        x === keys[0] ? 'O' : x[0].toUpperCase()
+      }]:${filters[x]} `
+    }
+  }
+
+  return formattedFilters
+}
```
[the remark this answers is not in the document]
nozomione: I've removed `getSortedFilters`, and instead, added `formatFacets` (which formats filter names in each facet) and used in `getFilterCombination` (at d365c507d1043e21b96038c9d7b5c8a9c87d4882).

The output looks good also. Thank you for this great suggestion 👍 !

src/analytics/helpers.js
```diff
@@ -0,0 +1,94 @@
+import moment from 'moment'
+import { links } from 'config'
+import formatFacetNames from 'helpers/formatFacetNames'
+import formatFilterName from 'helpers/formatFilterName'
+import getReadable from 'helpers/getReadable'
+
+const datasetOptionsKeys = ['aggregate_by', 'scale_by', 'quantile_normalize']
+
```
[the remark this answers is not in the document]
nozomione: I've made the following changes:

Added the following lists to `analytics/helpers`:
- `supportedEvents`: a list of supported custom events, which extends `genericEvents` (GA4 built-in events)
- `supportedDimensions`: a list of supported custom dimensions 

Added validation to `event()`:
- validation for supported event and dimension names
- throws an error if unsupported names are used

This is indeed a great replacement for the google sheet(custom event list) 👍 , see commit 32367c1b23106a3972f9aee2885524ba58f154c0.

src/analytics/gtag.js
nozomione: I originally borrowed the concept from the current `Refine.bio - GA4` in google analytics, which tracks API token creations by counting unique `token` values. I initially adapted this idea to count unique user dataset downloads. 

I've removed the `token` parameter and `trackDatasetDownload` now simply counts the dataset download as `dataset_download`, see at 3c77006020fef5ed3bef96e14baed8192e7d0743 👍

src/analytics/helpers.js
nozomione: Since we eliminated the use of template strings for dynamically assigning parameter names and now define them explicitly, the `getSnakeCase` method is no longer necessary. Thus, it has been removed, see at d365c507d1043e21b96038c9d7b5c8a9c87d4882.

nozomione:
I've updated this PR based on the suggestions and it's ready for your review. Thank you, David!

src/hooks/useSearchManager.js
```diff
@@ -2,6 +2,7 @@ import { useContext } from 'react'
 import { useRouter } from 'next/router'
 import { SearchManagerContext } from 'contexts/SearchManagerContext'
 import { options } from 'config'
+import formatFacetNames from 'helpers/formatFacetNames'
```
davidsmejia, line 5: This is unused so we can delete it now.

davidsmejia (commented):
Looking great just want to clean a couple things up.

davidsmejia (approved):
LGTM

src/analytics/helpers.js
```diff
@@ -0,0 +1,117 @@
...
+]
+export const event = (eventName, value = {}, nonInteraction = false) => {
+  try {
+    // throws an error if the given eventName or any dimension name is unsupported
+    const unsupportedDimensions = Object.keys(value).filter(
+      (key) => !supportedDimensions.includes(key)
+    )
+    if (unsupportedDimensions.length > 0) {
+      throw new Error(
+        `The dimension name${unsupportedDimensions.join()} is unsupported`
```
davidsmejia, line 61:
```suggestion
        `Unsupported dimension names: ${unsupportedDimensions.join(', ')}`
```

Issue #366, opened 2024-09-20
Expanding getReadable usage


nozomione:
### Context

This epic is for tracking the following tasks related to expanding the usage of `getReadable`:

- https://github.com/AlexsLemonade/refinebio-web/issues/367
- Clean up `config/options.compendia` property from config (3 stacked issues):
   - https://github.com/AlexsLemonade/refinebio-web/issues/376
   - https://github.com/AlexsLemonade/refinebio-web/issues/377
   - https://github.com/AlexsLemonade/refinebio-web/issues/378
   - https://github.com/AlexsLemonade/refinebio-web/issues/401

These issue PRs will be stacked and created in the order listed above.

(Other issues will be added later)

nozomione:
All issues listed in this epc have been merged into `dev`.

PR #365, opened 2024-09-18
359 - Handle FileSummary rendering using local state

nozomione:
## Issue Number

Epic: #358 

Closes #359

Merged into the feature branch: `feature/epic-refactor-dataset-page-data-handling`
Base & target branch: `nozomione/360-refactor-data-set-page-to-handle-dataset-id-via-url-query`

## Purpose/Implementation Notes
Please note that I've updated the issue context: This update also supports both the `/download` and `/dataset/id` routes, and the global context `regeneratedDataset` has been entirely removed from the codebase, as it has become obsolete and has been replaced by the local state. Thank you David!

I've made the following changes to support the File Summary section UI updates using the local state.

Changes includes:
- Removed the global context state `regeneratedDataset` from contexts/hooks
- Updated  the`useDatasetManager` hook methods based on the above changes (e.g., remove no longer required ID check etc)
- Updated the `DatasetRegenerate` component to take the entire data structure, i.e.,`dataset`(this component has no download options form)
- Updated the `FileSummary` component: 
  - Added a new local state `regeneratedDataset` for the UI swap in the File Summary section
  - Added a new component `RegenerateDatasetDownloadOptionsForm` as a wrapper component for `DownloadOptionsForm`, which was extracted from `FileSummary`
  - Added `helpers/getDownloadFilesData`, previously `downloadFilesData` extracted from `FileSummary`
 - Updated the shared `DownloadOptionsForm` component to support the `regenerateDataset` updates (in `dataset/id`), as well as My Dataset updates (in `download/`) on download options change 

Other change includes:
- Added a check for `regeneratedDataset` in `trackRegeneratedDataset` to handle cases when it is `undefined`  , preventing a type error in `getDatasetOptionsChanges` (only send this GA dimension when the download options changes)

## Types of changes

- Refactor (addresses code organization and design mentioned in corresponding issue)

## Functional tests

N/A

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

N/A

PR #369, opened 2024-09-26
367 - Add getReadableOptions helper for options

### LINKED ISSUE #367: Add getReadableOptions helper for options
### Context

Epic: #366

Target branch: `dev`

In PR #321, we introduced [`helpers/getReadable`](https://github.com/AlexsLemonade/refinebio-web/blob/dev/src/helpers/getReadable.js) and  [`config/readable`](https://github.com/AlexsLemonade/refinebio-web/blob/dev/src/config/readables.js). 

In this issue, we want to expand their usage across our codebase by adding a new helper function `getReadableOptions`.

This function will utilize `getReadable` to dynamically generate the appropriate data structure for rendering options (labels and values) in Grommet’s `Select` and `RadioButtonGroup` built-in components. This approach will reduce redundancy by eliminating the need for separate `options` definitions for each component and simplifying our implementation.

### Problem or idea

Currently, we explicitly define an array of `options` objects for each `Select` and `RadioButtonGroup`, and pass them directly to the components.

e.g.) The current approach for the aggregation select dropdown is as follows:
```js
//  The aggregation options explicitly defined in config/options
const options = { aggregation: [
    { label: 'Experiment', value: 'EXPERIMENT' },
    { label: 'Species', value: 'SPECIES' }
  ],
}

// AggregateOptions component
import { options } from 'config'; // import the options

export const AggregateOptions = () => (
  <Select
    labelKey="label"
    name={name}
    options={options.aggregation} // pass the options here
  />
);
```

e.g., ) The updated approach:
```js
<Select
  labelKey="label"
  name={name}
  options={getReadableOptions(['EXPERIMENT', 'SPECIES'])} // returns the generated options
/> 
```

This helper will take an array of option names and return the corresponding `options` for `Select` which matches with the `aggregation` data structure in `config/options`. 

### Solution or next step

1. Add a new helper function `getReadableOptions` that utilizes `getReadable` to return the correct options format for `Select` and `RadioButtonGroup`
2. Replace the passed `options` prop with `getReadableOptions` for `Select` and `RadioButtonGroup`
3. Update `config/readable` to support the new implementation and clean up properties for `Select` and `RadioButtonGroup` in `config/options`

nozomione:
## Issue Number

Epic: #366

Closes #367

Dependency (blocked by): #365

I've added the `getReadableOptions` helper and updated the implementations for Grommet `Select` and `ReadioButtonGroup` components, as well as cleaned up `config/options`.

Changes include:
- Added new helpers `getReadableOptions` and `uninqueArray` (for DRY)
- Added new properties to `config/readable` and cleaned up `config/options` (including the removal of duplicate definitions of `pageSizes` and  the `requestDataForm`)
- Updated the components to use `getReadableOptions`

Other changes:
- Added a type check to `gtag.TrackLink` to prevent en error (in 89dd3f386f6c262f7153f15dc1bc1acd66c0069b)

## Types of changes

## Functional tests
N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/src/analytics/gtag.js
+++ b/src/analytics/gtag.js
@@ -101,7 +101,7 @@ const trackExploredUsageClick = (link) => {
 }
 // tracks internal and external link clicks
 const trackLink = (link) => {
-  if (typeof link === 'string' && link.startsWith('http')) {
+  if (link.startsWith('http')) {
     trackExternalClick(link)
   } else {
     trackInternalClick(link)
--- a/src/analytics/gtag.js
+++ b/src/analytics/gtag.js
@@ -101,7 +101,7 @@ const trackExploredUsageClick = (link) => {
 }
 // tracks internal and external link clicks
 const trackLink = (link) => {
-  if (link.startsWith('http')) {
+  if (typeof link === 'string' && link.startsWith('http')) {
     trackExternalClick(link)
   } else {
     trackInternalClick(link)
--- a/src/config/options.js
+++ b/src/config/options.js
@@ -1,8 +1,4 @@
 export const options = {
-  aggregation: [
-    { label: 'Experiment', value: 'EXPERIMENT' },
-    { label: 'Species', value: 'SPECIES' }
-  ],
   // setting for the compendia
   compendia: {
     // default values for common queries
@@ -48,19 +44,6 @@ export const options = {
       ARRAY_EXPRESS: 'ArrayExpress'
     }
   },
-  // setting for the request data form
-  requestDataForm: {
-    radioPediatricCancer: [
-      { label: 'Yes', value: 'Yes' },
-      { label: 'No', value: 'No' }
-    ],
-    radioPrimaryApproach: [
-      { label: 'Bench Research', value: 'Bench Research' },
-      { label: 'Computational Research', value: 'Computational Research' },
-      { label: 'Clinical Research', value: 'Clinical Research' },
-      { label: 'AI/ML Research', value: 'AI/ML Research' }
-    ]
-  },
   // setting for the samples table
   samplesTable: {
     // default values for common queries
@@ -68,9 +51,10 @@ export const options = {
       offset: 0,
       limit: 10
     },
-    page: 1,
-    pageSizes: [10, 20, 50]
+    page: 1
   },
+  // setting for the page size
+  pageSizes: [10, 20, 50],
   // setting for the search manager
   search: {
     // default values for common queries
@@ -92,7 +76,6 @@ export const options = {
       platform_accession_codes: 'platform',
       technology: 'technology'
     },
-    pageSizes: [10, 20, 50],
     sortby: [
       {
         label: 'Best Match',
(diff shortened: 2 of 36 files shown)


src/config/options.js
```diff
@@ -116,6 +99,8 @@ export const options = {
       }
     ]
   },
+  // TODO: This will be cleaned up once the following PR is merged
+  // https://github.com/AlexsLemonade/refinebio-web/pull/365
```
nozomione, line 103: As noted here, `transformation` will be removed once #365 is merged (it has already been removed in that PR).

src/analytics/gtag.js
```diff
@@ -101,7 +101,7 @@ const trackExploredUsageClick = (link) => {
 }
 // tracks internal and external link clicks
 const trackLink = (link) => {
-  if (link.startsWith('http')) {
+  if (typeof link === 'string' && link.startsWith('http')) {
```
nozomione, line 104: I've added this type check to prevent an error encountered during the development, thank you!

src/config/options.js
```diff
@@ -48,29 +44,17 @@ export const options = {
...
     commonQueries: {
       offset: 0,
       limit: 10
     },
-    page: 1,
-    pageSizes: [10, 20, 50]
+    page: 1
   },
+  // setting for the page size
+  pageSizes: [10, 20, 50],
```
nozomione, line 57: Although this definition of the page size dropdown was identical, it was defined seperately in `search` and `samplesTable`, so I removed the duplicates.

src/config/options.js
```diff
@@ -48,29 +44,17 @@ export const options = {
...
-      { label: 'Yes', value: 'Yes' },
-      { label: 'No', value: 'No' }
-    ],
-    radioPrimaryApproach: [
-      { label: 'Bench Research', value: 'Bench Research' },
-      { label: 'Computational Research', value: 'Computational Research' },
-      { label: 'Clinical Research', value: 'Clinical Research' },
-      { label: 'AI/ML Research', value: 'AI/ML Research' }
-    ]
-  },
```
nozomione, line 63: I've cleaned up these definitions for the request data forms as both the labels and values are identical. They are no defined within the components and passed directly to Grommet components.

nozomione:
I've included the same change to `helpers/getRedable` aligning with the commit be8cc329255982685aa45b387f8c254429bdc895.

Additionally, I've applied 1:1 feedback to exclude `MoveToDatasetButton` from using `helpers/getReadableOptions`👍 

This PR is ready for another look, thank you David!

davidsmejia (approved):
One thing to address before merge and we can open a new issue to clean up how we pass data to child components.

src/analytics/gtag.js
nozomione: I mistakenly referred to `SearchCardFooter` instead of `SearchCardHeader`. Apologies for the confusion! I've now updated `SearchCardHeader` to ensure the `href` properly passes the formatted string and remove the type check for `link` from `gtag.trackLink`. Thank you David 👍!

nozomione:
I mistakenly referred to `SearchCardFooter` instead of `SearchCardHeader`. Apologies for the confusion! I’ve now updated `SearchCardHeader` to ensure the `href` prop properly passes the formatted string and removed the `link` parameter's type check in `gtag.trackLink` accordingly!

Now, I’m merging this PR into `dev`. Thank you, David! 👍

Issue #358, opened 2024-09-06
Refactor Dataset Page Data Handling


nozomione:
### Context
 
Branch: `feature/epic-refactor-dataset-page-data-handling`

This epic is to track the following tasks to enhance the dataset page:

1. https://github.com/AlexsLemonade/refinebio-web/issues/360
    - Update the dataset page to process the dataset ID via the URL query instead of dataset types
    - Move the JSX conditional rendering logic from the parent component to child components 
    - Include https://github.com/AlexsLemonade/refinebio-web/issues/362
2. https://github.com/AlexsLemonade/refinebio-web/issues/359 
    - Implement local state management within the `FileSummary` component to handle the file summary updates based on user-selected  regenerated dataset download options
3. https://github.com/AlexsLemonade/refinebio-web/issues/368 
    - Remove the `usePageRendered` hook dependency from the codebase 
4. https://github.com/AlexsLemonade/refinebio-web/issues/361
    - Extract the response formatting logic from the dataset manager hook and move it to data presentation components
5. https://github.com/AlexsLemonade/refinebio-web/issues/370
   - Add a helper to manage the dataset lifecycle state variables for the UI rendering

These issue PRs will be stacked and created in the order listed above.

(Other issues might be added later)

nozomione:
All issues listed in this epc have been merged into `dev`.

PR #373, opened 2024-09-27
370 -  Add getDatasetState helper for the dataset lifecycle

### LINKED ISSUE #370: Add getDatasetState helper for the dataset lifecycle
### Context

Epic: #358

Dependency: #361
A PR for this issue will be based off the above issue's PR branch.
(Due to frequent edits to the same files, the above dependency was added to this issue to minimize merge conflicts.)

In `/download` and `dataset` routes, we render the UI based on the dataset lifecycle states which include unprocessed, processing, processed (unexpired or expired), and processing error. 

Currently, we set these values by destructuring `dataset` and creating all the appropriate variables within the components for conditional UI rendering logic.

### Problem or idea

This approach clutters each component's implementation. Thus, we should create a dedicated helper to handle the assignments of these dataset lifecycle states. This will improve the readability and maintainability of our codebase.

### Solution or next step

- Add a new helper `getDatasetState` that does the following:
  - Takes `dataset` as its argument
  - Returns the following lifecycle variables (names may differ): 
    -  (Mutable)
       - **unprocessed** : `isNotProcessed`
    - (Immutable) 
      - **processing** : `isProcessing`
      - **processed**
        - **unexpired** (available for download) : `isReady`
        - **expired** (available for regeneration download) : `isRegenerative`
      - **processing error** : `isProcessingError`
- Use this new helper and clean up locally defined dataset lifecycle state variables and destructuring of `dataset` from the components

nozomione:
## Issue Number

Epic: #358 

Closes #370

Merged into the feature branch: `feature/epic-refactor-dataset-page-data-handling`
Base & target branch: `nozomione/361-decouple-format-experiments-from-dataset-manager`

I've add `helpers/getDatasetState` to facilitate UI rendering and replaced the local variables / destructuring of `dataset` from the components with this new helper.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


src/helpers/getDatasetState.js
```diff
@@ -0,0 +1,25 @@
...
+    expires_on: expiredOn,
+    is_available: isAvailable,
+    is_processed: isProcessed,
+    is_processing: isProcessing,
+    success
+  } = dataset
+
+  const isExpired = moment(expiredOn).isBefore(Date.now())
+  const isComplete = isProcessed && isAvailable && success // successfully processed
+  const isProcessingError = success === false // failed during processing
```
nozomione, line 14: I am thinking about renaming this to `hasProcessingError`. I originally matched it with other property names (using the `is` prefix), but I think renaming might be a better fit!

davidsmejia: i think `isFailed` should work here

src/components/Download/DownloadOptionsForm/index.js
```diff
@@ -21,22 +22,24 @@ export const DownloadOptionsForm = ({
   const { updateDataset, getDownloadOptions, updateDownloadOptions } =
     useDatasetManager()
   const { setResponsive } = useResponsive()
-  const isProcessed = dataset?.is_processed
+  const [isRegenerate, setIsRegenerate] = useState(false)
```
davidsmejia, line 25: you can use `canRegenerate` here if you want to capture potential state instead of current state.

davidsmejia (commented):
Just a couple suggestions

src/helpers/getDatasetState.js
nozomione: Great name 👍

src/components/Download/DownloadOptionsForm/index.js
nozomione: This is better 👍!

nozomione:
I've applied your suggestions and this PR is ready for your review. Thank you David!

davidsmejia (approved):
LGTM

PR #400, opened 2024-10-25
378 - Use query.type for component props and update data fetching for Compendia

### LINKED ISSUE #378: Use query.type for component props and update data fetching for Compendia
### Context
Epic: #366

In #376 (**stacked issue 1**),  we removed the `options` dependency and incorporated the usage of `getReadable` in the compendia components. 

In  #377 (**stacked issue 2**), we also utilize the `type` query to set the current compendia tab in the `[type] ` page component and its sub-component `Tabs`.

Currently, we use the `type` props in the sub-components of `Tabs` to determine the rendering of each tab's contents. 

In this **stacked issue 3**,  we want to further optimize those sub-components by utilizing  the `type` query as component props and remove the `type` props from the tab components.

Additionally, we want to change the data fetching location as part of this process. 

Currently, the shared `Download` component handles the data fetching. We want to shift this responsibility to its parent component (i.e,. the `[type]` page component), which will then pass the fetched data to to its sub-components including `Download`.

### Problem or idea

To achieve this, we should implement the following changes:
- Utilize the `type` query value as the component's prop value
- Fetch the `compendia` data in each tab component (parent) instead of the shared sub-component
- Pass the entire fetched data structure to the sub-component

**NOTE:** In the current production environment of refine.bio, the UI fetches data for each tab every time a tab is switched. We will adjust the approach to fetch all necessary data (both normalized and rna-seq) at the top level only once on page load. This change will reduce API calls and enhance overall UI performance.

### Solution or next step

1. Replace the `type` props with the `type` query
2. Remove the `type` props from the tab-related sub-components
3. Make the API call from the `[type]` page component on page load and pass the fetched `compendia` to its sub-components
4. Remove the API call from the`Download` component
5. Adjust  the`useCompendia` hook and the codebase accordingly

nozomione:
## Issue Number

Epic: #366

Closes #378 (Stacked issue 3)

Base PR branch: #399

I've updated the implementation to utilize `query.type` and removed the `type` from the tab-related sub-components. Additionally, I've moved the data fetching location from `Download` (on every tab switch - each tab fetches its data separately) to `[type]` (only fetch once on page load) to reduce API calls and improve UI performance. 

Changes include: 
- Used the `type` query and cleaned up the `type` props from the tab-related sub-components
- Fetching the `compendia` in `[type]` and remove the API call from `Download`
- Updated the `useCompendia` hook

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/src/components/Compendia/NormalizedTab.js
+++ b/src/components/Compendia/NormalizedTab.js
@@ -3,16 +3,15 @@ import { Box, Heading, Text } from 'grommet'
 import { useResponsive } from 'hooks/useResponsive'
 import { Button } from 'components/shared/Button'
 import { FixedContainer } from 'components/shared/FixedContainer'
-import { links, options } from 'config'
+import { links } from 'config'
 import { DownloadBlock } from './DownloadBlock'
 
-export const NormalizedTab = ({ type = 'normalized' }) => {
+export const NormalizedTab = ({ compendia }) => {
   const { setResponsive } = useResponsive()
-  const tabName = options.compendia.heading[type]
 
   return (
     <Box animation={{ type: 'fadeIn', duration: 350, delay: 200 }}>
-      <DownloadBlock type={type} />
+      <DownloadBlock compendia={compendia} />
       <FixedContainer
         pad={{
           horizontal: setResponsive('large', 'medium', 'basex15'),
@@ -38,7 +37,7 @@ export const NormalizedTab = ({ type = 'normalized' }) => {
             quantile normalization.
           </Heading>
           <Button
-            aria-label={`Go to the refinebio docs - ${tabName}`}
+            aria-label="Go to the refinebio docs - Normalized Compendia"
             href={links.refinebio_docs_normalized_compendia}
             label="Learn More"
             secondary
--- a/src/components/Compendia/RNASeqTab.js
+++ b/src/components/Compendia/RNASeqTab.js
@@ -1,11 +1,12 @@
 import { memo } from 'react'
 import { Anchor, Box, Heading, Paragraph, Text } from 'grommet'
 import { useResponsive } from 'hooks/useResponsive'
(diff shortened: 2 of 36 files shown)


src/components/Compendia/NormalizedTab.js
```diff
@@ -38,7 +37,7 @@ export const NormalizedTab = ({ type = 'normalized' }) => {
             quantile normalization.
           </Heading>
           <Button
-            aria-label={`Go to the refinebio docs - ${tabName}`}
+            aria-label="Go to the refinebio docs - Normalized Compendia"
```
nozomione, line 40: Since the `aria-lable` value is static (specific to the nomalized tab), I've removed `getReadable` and hard-coded the value here.

src/components/Compendia/RNASeqTab.js
```diff
@@ -76,9 +75,7 @@ export const RNASeqTab = ({ type = 'rnaSeq' }) => {
               Note: This compendia is not normalized or aggregated.
             </Paragraph>
             <Button
-              aria-label={`Go to the refinebio docs - ${getReadable(
-                'rna-seq'
-              )}`}
+              aria-label="Go to the refinebio docs - RNA-seq Sample Compendia"
```
nozomione, line 78: Same as above.

src/config/options.js
```diff
@@ -8,11 +8,11 @@ export const options = {
     },
     heading: {
       normalized: 'Normalized Compendia',
-      rnaSeq: 'RNA-seq Sample Compendia'
+      'rna-seq': 'RNA-seq Sample Compendia'
```
nozomione, line 11: This is a temporary fix to prevent an error due to removal of the `type` prop. The `compendia` property will be removed from `config/options` in a subsequent issue (#401).

davidsmejia (approved):

PR #410, opened 2024-11-10
407 - Validate tokens with API call instead of regex in useToken

### LINKED ISSUE #407: Validate tokens with API call instead of regex in useToken
### Context

Currently, we validate the token using a regex in the `validateToken` method of the `useToken` hook. We want to replace this with an actual API call.

### Problem or idea
We should call the `token` endpoint from `validateToken` to verify the token's validity on the server.

### Solution or next step
Replace the regex check in `validateToken` with an API call

nozomione:
## Issue Number

Closes #407

I've updated the `validateToken` method in the `useToken` hook to make an API call and adjusted the codebase accordingly. 

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/src/components/Dataset/DatasetReady.js
+++ b/src/components/Dataset/DatasetReady.js
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { Box, Heading, Text } from 'grommet'
 import gtag from 'analytics/gtag'
 import { useDatasetManager } from 'hooks/useDatasetManager'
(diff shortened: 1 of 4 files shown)


src/components/Dataset/DatasetReady.js
```diff
@@ -17,14 +17,23 @@ export const DatasetReady = ({ dataset }) => {
   const { downloadDataset } = useDatasetManager()
   const { setResponsive } = useResponsive()
   const { validateToken } = useToken()
-  const hasToken = validateToken()
-  const [termsOfUse, setTermsOfUse] = useState(hasToken)
+  const [acceptedTerms, setAcceptedTerms] = useState(false)
```
davidsmejia, line 20: I don't think this should be defined local to the component. This should be in `useToken`. It should start false before being queried, then updated when validated.

davidsmejia (commented):
Take a look at your usage of tokens and let me know if you have a suggestion for what we can do to `useToken` to make it more usable.

src/components/Dataset/DatasetReady.js
nozomione: We're removing `useToken` and moving user token validation to the refinebio context during application load, rather than at the time of file downloads (#413). As a result, these checks are no longer necessary and will be removed from the components including this one.

davidsmejia (approved):

PR #419, opened 2024-11-26
389 - Reimplement useSamplesTableContext to useSamplesContext

### LINKED ISSUE #389: Rename samplesTable to samplesQuery in SamplesTableManagerContext
### Context

Epic: #388 

Feature branch: `feature/epic-refacter-samples-table-hook-and-components`

In `SamplesTableManagerContext`, we currently have a context state called `samplesTable` which stores the queries needed to fetch samples table data from the `samples` endpoint. 

We want to rename this state to `samplesQuery` to align with the endpoint name.

### Problem or idea

We should rename `samplesTable` to `samplesQuery` in `SamplesTableManagerContext` and update the codebase accordingly.  

### Solution or next step

1. Rename `samplesTable` to `samplesQuery` in `SamplesTableManagerContext`
2. Update the state name to `samplesQuery` in `useSamplesTableManager` and all components that utilize it

nozomione:
## Issue Number

Closes #389

I've re-implemented `SamplesTableContext` and its hook `useSamplesTableContext` as `SamplesContext` and `useSamplesContext` respectively.

Changes include:
- Rename the previous context and hook as `SamplesContext` and `useSamplesContext`, and adjust the imports 
- Replaced the term `table` in all methods, variables, and states (e.g., `tableData` renamed to `samples`)
- Remove `config/options.SamplesTable` (no longer needed)
- In the `useSamplesContext` hook:
   - Updated `samplesQuery` by removing client-specific query names:
      - `page` renamed to `offset`
      - `pageSize` renamed to `limit`
      - `sortBy` renamed to `ordering`
      - Added the `initialQueries` parameter (required) to the hook for consumers
      -  Improved the data fetching logic in `useEffect` to optimize API calls on `samplesQuery` changes
- In the `SamplesTable` component: 
  - Defined the initial queries to be passed to useSamplesContext (no longer using `config/options`)
- Updated the `ShowOnlyAddedSamplesFilter` component to re-fetch the data using `useEffect` (no longer used handle method)

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


src/hooks/useSamplesContext.js
```diff
@@ -0,0 +1,130 @@
...
+
+  /* Common */
+  const getSamples = async () => {
+    const { offset, limit, datasetId, filterBy, ordering } = samplesQuery
+    const params = {
+      ...initialQueries,
+      offset,
+      limit,
+      ...(datasetId && { dataset_id: datasetId }),
+      ...(filterBy && { filter_by: filterBy }),
```
davidsmejia, line 35: No need to rename these since they are for the API and inside an object so we should just spread samplesQuery with initialQuery into a new object.

nozomione:
I've applied the feedback and this PR is now ready for your review. Thank you David!

src/hooks/useSamplesContext.js
nozomione: Here, to avoid using the logical AND `...(propName && { ...`, I also added a new `filterNullAndEmptyFromObject` helper to remove `null` and empty string values from `params` and spread `samplesQuery` directly.

src/hooks/useSamplesContext.js
```diff
@@ -0,0 +1,127 @@
...
+
+  return {
+    samples,
+    samplesQuery,
+    hasError,
+    hasSamples,
+    loading,
+    totalSamples,
+    getSamples,
+    updateFilterBy,
```
davidsmejia, line 119:
```suggestion
    setFilterBy,
    filterBy: samplesQuery.filter_by,
```

We should update this to match existing conventions

nozomione: **NOTE:** We decided not to update the method names at this moment and will revisit them in the future if necessary.

nozomione:
I've applied the feedback and this PR is ready for your review. Thank you David!

PR #421, opened 2024-12-02
413 - Update token validation flow

### LINKED ISSUE #413: Update token validation flow
### Context

Currently, the user token is validated only when downloading a file via the `useToken` hook. We want to change this approach to validate the token during the initial load of the application.

### Solution or next step

To achieve this: 
1. Add a new state `acceptedTerms` stored in localStorage:
    - When this is `true`, create a token or regenerate it (for to handle API version changes) via a side effect
2. Validate the user token during the application's initial load
3. Remove user token validation from other parts of the codebase

nozomione:
## Issue Number

Closes #413

I've implemented the user token validation flow during the initial load of the application, which replaces the `useToken` hook, and adjusted the codebase accordingly. 
 
Changes include:
- Added a new `acceptedTerms`context state to track whether the user has previously accepted the terms, which is used to determine if validation is necessary during the application load
- Removed the token validation check from the codebase
- Adjusted the visibility of the terms of use checkbox in the download forms

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A


nozomione:
I've updated the PR based on the 1:1 feedback, and it's ready for another review. Thank you, David!

Changes include:
- In `RefinebioContext`, added:
  - The application-level token on initial load
  - A side effect for syncing the token state with `acceptTerms`
- Adjusted the consumer hooks (removed token generation i.e., `createToken`) and components (updated form submission and visibility of the terms of use checkbox)
- Renamed the helper and interface method parameter from `params` to `body`

davidsmejia (commented):
Two changes for now:
- only use token as a dependency of useEffect
- acceptedTerms should be used for component logic
- use `setAcceptedTerms` and remove `applyAcceptedTerms`

nozomione:
I've applied the following changes:

>     * only use token as a dependency of useEffect
>     * acceptedTerms should be used for component logic
>     * use `setAcceptedTerms` and remove `applyAcceptedTerms`

To ensure the `token` state is used only within hooks and not in components, I've created a temporary `useTriggerSubmit` hook that triggers the `submit` method in each consumer component when both `token` and the required condition (e.g., download options) are ready.

nozomione:
David, thank you for this great idea for the `useWaitFor` hook! 🚀

We discussed using the `useWaitFor` hook to ensure that a callback is only called once a specified condition is met. In our case, certain async methods need to be executed only after a valid token is ready (which is fetched asynchronously via `RefinebioContext`).

```js
import { useEffect, useRef } from 'react'

export const useWaitFor = (condition) => {
  const resolveRef = useRef()
  const promiseRef = useRef(new Promise((resolve) => {
    resolveRef.current = resolve
  }));

  useEffect(() => {
    if (condition) resolveRef.current() // resolve only when condition is met
  }, [condition])
  
  // once the promise is successfully resolved, call the callback
  return (cb) => async (...args) => promiseRef.current.then(cb(...args))
}
```

**Usage:**
1. Instantiate the hook at the `RefinebioContext` provider level and define the `waitForToken` method:
```js
const waitForToken = useWaitFor(validToken) 
```
2. Wrap any methods that require a valid token with `waitForToken` 
```js
const downloadDataset = waitForToken(async (...) => {...}) 
```

Based on this insight, I've updated the codebase. I also customized the hook to initialize only once (for re-renders). Additionally, the hook now passes the latest condition to the callback as the last parameter after resolution, and I've adjusted the method signatures accordingly.

This PR is ready for another look. Thank you David!

nozomione:
I've applied the 1:1 feedback, replaced `useWaitFor` with `usePromise`, and made the following adjustments to the codebase:
- Handled form submission using side effects within the components
- Threw an error if `acceptedTerms` is `false` in the following `useDatasetManager` methods:
   - `downloadDataset`
   - `startProcessingDataset`
- Removed the method definitions for creating and activating tokens, and handled them directly within `useEffect` blocks in `RefinebioContextProvider`

This PR is ready for your review. Thank you, David!

davidsmejia (commented):
This is almost there, just left a few comments.

davidsmejia (approved):
LGTM

PR #435, opened 2025-01-06
386 - Refactor MoveToDatasetButton / MoveToDatasetModal components

### LINKED ISSUE #386: Refactor MoveToDatasetButton / MoveToDatasetModal components
### Context

In the `dataset` route, we display the "Move to Dataset" button (`MoveToDatasetButton`) which allows users to replace or add shared dataset samples to their “My Dataset”.

move-to-dataset-button

When a user clicks `MoveToDatasetButton`, we currently determine the button behavior using the `handleMoveToDataset` method:

- If any samples are present, the button opens `MoveToDatasetModal` that lets users to replace or add shared dataset samples to their "My Dataset".
- If no samples are present, we directly append the shared dataset samples to the user’s “My Dataset” without opening `MoveToDatasetModal`.

### Problem or idea

Instead of determining the button behavior using `handleMoveToDataset`, we want to utilize an early return to render different buttons in the UI based on the total sample count.

Additionally, we want to move some of the variables from `MoveToDatasetButton` to `MoveToDatasetModal` during this refactoring process.

```js
 const handleMoveToDataset = async () => {
    if (totalSamples > 0) {
      openModal(id)
    } else {
      await addSamples(selectedDataset.data)
      ...
  }
```

e.g., New approach:

```js

// Render a button to append shared dataset samples when no samples
if (totalSamples === 0) {
    return (
       <Button 
          label="Move to Dataset"
          secondary
          responsive
          onClick={handleMoveToDataset}
    />)
}

// Otherwise, render a button to open the modal
return (
    <Modal
      button={
        <Button
          label="Move to Dataset"
          secondary
          responsive
         onClick={openModal}
        />
      }
    >
    ...
  )
  ```

### Solution or next step

- Use the early return in `MoveToDatasetButton` to conditionally render:
  - A button for opening `MoveToDatasetModal` when samples are present
  - A button for adding the shared dataset directly when there are no samples
- Move the following variables from `MoveToDatasetButton` to `MoveToDatasetModal`:
  - `radioOptions`
  - `defaultValue`
  - `value`/`setValue`

### PR DESCRIPTION
**Pull request by nozomione**
## Issue Number

Closes #386 

Base branch: PR #421

I've used an early return to handle the case of no samples in My Dataset, and optimized the action methods (append, replace) of the "Move to Dataset" button.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### CODE REVIEW

File: src/components/Dataset/MoveToDatasetButton.js [hunk 1]
```diff
@@ -1,74 +1,69 @@
...
-    }
+  const handleAppend = async () => {
+    await addSamples(dataset.data)
+    getRedirect('Appended')
+  }
+
+  const handleReplace = async () => {
+    await replaceSamples(dataset.data)
+    getRedirect('Moved')
+  }
```
davidsmejia, line 50:
```suggestion
const handlers = {
  append: handleAppend,
  replace: async () => {
    await replaceSamples(dataset.data)
    downloadRedirectWithPrefix('Moved')
  }
}

const handleSelectAction = (action) => {
   await handlers[action]()
   // analytics
   // close modal
}
```

davidsmejia (commented):
I think we can move some logic out of the modal and just have it take a callback for when a value is selected or cancelled.

src/components/Dataset/MoveToDatasetButton.js
nozomione: I've defined the `handlers` map for the actions and moved the `action` state to the parent component  `MoveToDatasetButton`.

Additionally, I've used  the `<Form>` element in `MoveToDatasetModal` to handle submit and reset events, and updated the props to include `onChange`, `onReset`, and `onSubmit` for better clarity.  

davidsmejia (approved):
LGTM

PR #433, opened 2025-01-03
427 - Rename SearchManagerContext state to searchParams and simplify useSearchManager implementation

nozomione:
## Issue Number

Closes #427

Base branch: PR #430

## Purpose/Implementation Notes

Changes include:
- Updated `SearchManagerContext`:
  - Renamed the `search` state to `searchParams`
  - Deleted the `filterOrders` state 
- Optimized the `useSearchManager` hook:
  - Replaced the `updateSearchQuery` method with a new `useSyncSearch` hook, which syncs the latest search parameters with the search results page's URL 
  - Replaced the `filterOrders` state with a local variable `filterOrders` in the `toggleFilter` method and deleted `addFilter`, `removeFilter`, `updateFilterOrders` methods that were used for this replaced state
  - Adjusted `helpers/getSearchQueryForAPI` to include the parameter values when defined
- Decouple the experiment page's `ref` query (used to indicate it's from the search results) from the `useSearchManager` hook in the following components:
  - `SearchCardHeader`
  - `SearchCardFooter`
- Adjusted the `pages/search` accordingly 
## Types of changes

- Refactor (addresses code organization and design mentioned in corresponding issue)

## Functional tests

N/A

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots

N/A

PR #443, opened 2025-01-17
439 - Refactor the SearchBox component to manage input reset with local state 

### LINKED ISSUE #439: Refactor the SearchBox component to manage input reset with local state
### Context

The current implementation of the shared `SearchBox` component relies on parent components to manage the search input field value via the `onChange` prop. This also includes the responsibility of clearing the input (i.e., resetting it to an empty string).

`SearchBox` is shared across the following components: 
- `pages/search/index`
- `Compendia/DownloadBlock`
- `Home/HeroBody`
- `SearchResults/SearchFilter`

In this PR, we'll shift the responsibility of resetting the input value to `SearchBox` itself by using local state. This change will simplify the state management of the input value within `SearchBox` and reduce the dependency on parent components.

Additionally, any event handler props passed from the parent (e.g., `onChange`, `onSubmit`) will be converted to no-op functions by default unless explicitly needed.

### Solution or next step

- Add a new local state to `SearchBox` for managing the input rest
- Set all event handler props to noop by default
- Update the parent components accordingly

nozomione:
## Issue Number

Closes #439

Base branch: PR #433

I've refactored the shared`SearchBox` component as follows and adjusted the components that utilize it:
- It handles clearing input field value using its local state
- It accepts event handler props as optional

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

N/A

### DIFF
--- a/src/components/Compendia/DownloadBlockForm.js
+++ b/src/components/Compendia/DownloadBlockForm.js
@@ -121,7 +121,7 @@ export const DownloadBlockForm = () => {
             reverse={false}
             responsive
             value={userInput}
-            onChange={(e) => setUserInput(e.target.value)}
+            onChange={setUserInput}
           />
           {showOptions && filteredOptions.length > 0 && (
             <Box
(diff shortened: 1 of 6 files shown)


davidsmejia (approved):
LGTM

PR #545, opened 2025-04-28
543 - Fix duplicate dataset IDs when regenerating a datase

### LINKED ISSUE #543: Fix duplicate dataset IDs when regenerating a dataset
### Context

Currently, when attempting to regenerate an expired dataset using the download options, and then processing that dataset via the request processing, a dataset ID duplication occurs.

- **Step 1:** Download Options Form generates the first dataset ID, which includes the download options set by the user (e.g., aggregated by `Species`, transformation set to `Zero to One`)

- **Step 2:** Request Processing Form generates the second dataset ID, which has the default download options defined by the BE API (i.e., aggregated by `Experiment`, transformation set to `None`), and this newly created ID is used for processing the dataset file download.

This results in a downloadable file being generated for regenerated dataset with the default download options.

### Problem or idea

We need to fix the dataset regeneration process to properly handle datasets with download options defined by users.

### Solution or next step

Identify and fix the bug in the dataset regeneration process to prevent duplicate dataset ID generation.

nozomione:
## Issue Number

Closes #543

I've fixed the bug in the dataset regeneration process and prevented duplicate dataset ID generation.

## Types of changes

## Functional tests

N/A

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Screenshots

N/A

### DIFF
--- a/src/hooks/useDatasetManager.js
+++ b/src/hooks/useDatasetManager.js
@@ -144,7 +144,7 @@ export const useDatasetManager = () => {
 
     const processingDatasetId = isMyDatasetId(id)
       ? myDatasetId
-      : await createDataset() // creates new dataset ID for one-off download and shared dataset
+      : id || (await createDataset()) // creates new dataset ID for one-off download and shared dataset
 
     const response = await updateDataset(processingDatasetId, body)
 
--- a/src/hooks/useDatasetManager.js
+++ b/src/hooks/useDatasetManager.js
@@ -144,7 +144,7 @@ export const useDatasetManager = () => {
 
     const processingDatasetId = isMyDatasetId(id)
       ? myDatasetId
-      : id || (await createDataset()) // creates new dataset ID for one-off download and shared dataset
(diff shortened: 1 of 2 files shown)


src/hooks/useDatasetManager.js
```diff
@@ -144,7 +144,7 @@ export const useDatasetManager = () => {
 
     const processingDatasetId = isMyDatasetId(id)
       ? myDatasetId
-      : await createDataset() // creates new dataset ID for one-off download and shared dataset
+      : id || (await createDataset()) // creates a new dataset ID for one-off download or shared dataset
```
davidsmejia, line 147:
```
    const processingDatasetId = isMyDatasetId(id)
      ? id
      : id || (await createDataset())

const processingDatasetId = id || (await createDataset())
```

Isn't this the same?

nozomione: Yes - cleaned up the check. The `dataset` prop is now passed to the form component, so there's no need to use the `MyDatasetId` value from localStorage here 👍

davidsmejia (approved):
LGTM

PR #548, opened 2025-05-02
536 - Remove Alternate Accession IDs links for E-GEOD

### LINKED ISSUE #536: Remove Alternate Accession IDs links for E-GEOD
### Context

It seems that pages for accession codes starting with `E-GEOD` are no longer available on the [EMBL-EBI](https://www.ebi.ac.uk/) website and they return a `404` error. However, the client still renders links for the `Alternative Accession IDs` for `E-GEOD`.

e.g. ) The link for `Alternate Accession IDs` [E-GEOD-73072](https://www.ebi.ac.uk/arrayexpress/experiments/E-GEOD-73072) returns a `404`:
https://www.refine.bio/experiments/GSE73072/host-gene-expression-signatures-of-h1n1-h3n2-hrv-rsv-virus-infection-in-adults

### Problem or idea

We should update the client and unlink the `E-GEOD` alternative accession IDs, as they are no longer available.

@deepa, let me know if removing the URLs for the `E-GEOD` alternative accession IDs is sufficient, or if we should completely hide the `Alternative Accession IDs` values by removing the row from the page (since the row is not displayed when there are no values).

Also tagging @davidsmejia for any additional feedback. 

e.g. ) Remove URLs (unlinked them but still display the text value):
Image

### Solution or next step

- [x] Get feedback from Deepa 
- [x] Implement the chosen solution

nozomione:
## Issue Number

Closes #536

I've unlinked the alternative accession IDs for `E-GEOD*`.  

Please see the latest preview [here](https://refinebio-gid8x9q45-ccdl.vercel.app/search)

## Types of changes

## Functional tests

N/A

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Screenshots

N/A

### DIFF
--- a/src/components/ExperimentCardBody.js
+++ b/src/components/ExperimentCardBody.js
@@ -19,7 +19,7 @@ export const ExperimentCardBody = ({ experiment, charLimit = 300 }) => {
   const [toggleDesciption, setToggleDescription] = useState(
     description.length > charLimit
   )
-  const isArrayExpress =
+  const isGEOD =
     alternateAccessionCode && alternateAccessionCode.startsWith('E-GEOD')
 
   return (
@@ -97,7 +97,7 @@ export const ExperimentCardBody = ({ experiment, charLimit = 300 }) => {
           Alternate Accession IDs
         </Heading>
         {alternateAccessionCode ? (
-          isArrayExpress ? (
+          isGEOD ? (
             <TextHighlight>{alternateAccessionCode}</TextHighlight>
           ) : (
             <Anchor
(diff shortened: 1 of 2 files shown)


src/components/ExperimentCardBody.js
```diff
@@ -18,6 +19,8 @@ export const ExperimentCardBody = ({ experiment, charLimit = 300 }) => {
   const [toggleDesciption, setToggleDescription] = useState(
     description.length > charLimit
   )
+  const isArrayExpress =
```
davidsmejia, line 22:
```suggestion
  const isGEOD =
```

davidsmejia (approved):
LGTM just the variable name should probably be more specific

